PR #27, opened 2017-11-10
Switch expected test entry structure to colocated, glob-based pattern

rtsao:
This change enables unit tests to be colocated with code, rather than enforcing a top-level tests directory.

Resolves https://github.com/fusionjs/fusion-cli/issues/28


build/compiler.js
```diff
@@ -50,11 +68,15 @@ function getConfig({target, env, dir, watch, cover}) {
   if (!fs.existsSync(path.resolve(dir))) {
     throw new Error(`Project directory must contain a ${main} file`);
   }
-  if (env === 'test' && !fs.existsSync(path.resolve(dir, serverTestEntry))) {
-    throw new Error(`Testing requires a ${serverTestEntry} directory`);
+  if (env === 'test' && !globby.sync(serverTestGlobs).length) {
+    throw new Error(
+      `Testing requires server tests in __tests__ or __tests__/__browser__`
```
nadiia: `__node__` instead of `__browser__`

PR #30, opened 2017-11-11
Handle errors in initial compilation in dev command

### LINKED ISSUE #29: Don't crash dev command on initial compilation errors
The `dev` command currently will crash if there are compiler errors on the initial compilation. We should make sure we are catching and reporting these errors, and keeping a watcher alive.

ganemone:
Fixes #29


ganemone:
!merge

PR #37, opened 2017-11-29
Add Flow step to pipeline.yml

AlexMSmithCA:
Created with fusion-orchestrate.


AlexMSmithCA:
Update:
- Some minor Flow annotation changes to fix check complaints.
- Using comment-syntax here to avoid breaking tests - which don't appear to transform away Flow syntax

rtsao:
!merge

PR #48, opened 2017-12-01
Include shared entry file in tests for real coverage

dennisgl:
A common issue with [nyc/istanbul](https://github.com/istanbuljs/nyc) is the coverage only covers as far as the test entry points reach but not the whole bundle, giving a false sense of high coverage. Here's a similar issue describing the problem: gotwarlost/istanbul#112

`--include-all-sources` in `Istanbul v1`, now `--all` in `Istanbul v2/nyc` extends the coverage area to the whole bundle. However, our bundles including test bundles are tree-shaked, therefore the option won't save the situation.

The changes here include shared entry point(currently `src/main.js` to test entry points passing to [multi-entry-loader](https://www.npmjs.com/package/multi-entry-loader)

Open for comments, given this basically eliminates tree-shaking for test bundles.

### DIFF
--- a/build/compiler.js
+++ b/build/compiler.js
@@ -53,7 +53,7 @@ function getConfig({target, env, dir, watch, cover}) {
   if (env !== 'production' && env !== 'development' && env !== 'test') {
     throw new Error('Invalid name: must be `production`, `dev`, or `test`');
   }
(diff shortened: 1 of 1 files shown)


dennisgl:
!merge

PR #55, opened 2017-12-08
Expose CLI options correctly

### LINKED ISSUE #57: Annotate CLI commands correctly in yargs
yargs help is currently incomplete, given the arguments that are actually used by the handler functions. Unneeded options should be removed, options of the form `no-something` should be handled correctly, and other valid options should appear in `--help`

lhorie:
Fixes some issues with some CLI command options, and improves contextual help from yargs

Resolves #57


lhorie:
!merge

PR #59, opened 2017-12-08
Remove `start --environment=test` in favor of `test --skip-build`

### LINKED ISSUE #58: Remove redundant CLI command: `start --environment=test`
`fusion start --environment=test` is equivalent to `fusion test --skip-build`. We should remove it to simplify the CLI API. (Also, it doesn't actually work)

lhorie:
Removes the command `start --environment=test` in favor of `test --skip-build`. The former is redundant, confusing and doesn't actually work.

Resolves #58


lhorie:
Not sure why codecov is whining. My diff doesn't add any code branches, as far as I can tell

lhorie:
!merge

PR #61, opened 2017-12-08
Support overriding the node environment in dev and test

### LINKED ISSUE #60: Support overriding the node environment in dev and test
It is a much better experience to have NODE_ENV overridden 
when running `fusion dev` and `fusion test` than throwing.
In production, we should still error on the side of caution
and throw, however in dev/test it should suffice to log a warning

ganemone:
Fixes #60

### DIFF
--- a/build/compiler.js
+++ b/build/compiler.js
@@ -105,6 +105,19 @@ function getConfig({target, env, dir, watch, cover}) {
   // 'test' and 'production' entries should both map to NODE_ENV='production'
   const nodeEnv = env === 'development' ? 'development' : 'production';
 
+  // Allow overrides with a warning for `dev` and `test` commands. In production builds, throw if NODE_ENV is not `production`.
+  const nodeEnvBanner =
+    `if(process.env.NODE_ENV && process.env.NODE_ENV !== '${nodeEnv}') {` +
+    `if ('${env}' === 'production') {` +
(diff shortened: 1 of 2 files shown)


test/cli/dev.js
```diff
@@ -13,7 +13,7 @@ test('`fusion dev` works', async t => {
   const entry = path.resolve(dir, entryPath);
 
   const {stop} = await run(
-    `dev --dir=${dir} --no-open=true --port=${await getPort()}`
+    `dev --dir=${dir} --no-open --port=${await getPort()}`
```
davidkearns: Unrelated change?

build/compiler.js
```diff
@@ -105,6 +105,19 @@ function getConfig({target, env, dir, watch, cover}) {
   // 'test' and 'production' entries should both map to NODE_ENV='production'
   const nodeEnv = env === 'development' ? 'development' : 'production';
 
+  // Allow overrides with a warning for `dev` and `test` commands. In production builds, throw if NODE_ENV is not `production`.
+  const nodeEnvBanner =
+    `if(process.env.NODE_ENV && process.env.NODE_ENV !== '${nodeEnv}') {` +
+    `if ('${env}' === 'production') {` +
```
lhorie: Better to do this check at compile time rather than bundling the checks into the runtime

davidkearns: The logic here is sound, but I wonder if there's any (minuscule) benefit of pulling this particular `if/else` out of the template since we can resolve these values earlier at build time.

ganemone:
!merge

PR #64, opened 2017-12-08
Ensure `test --watch` continues watching after test failure

### LINKED ISSUE #51: fusion test --watch exits when tests fail
This should behave the same way that `dev` does. On test failure, we should wait for a recompilation and then re-run tests.

lhorie:
Resolves #51

### DIFF
--- a/commands/test.js
+++ b/commands/test.js
@@ -64,7 +64,9 @@ exports.run = async function({dir = '.', cover, watch, skipBuild}) {
   await runTests();
 
   if (watch) {
-    compiler.on('done', runTests);
+    compiler.on('done', () => {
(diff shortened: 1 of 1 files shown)


commands/test.js
```diff
@@ -64,7 +64,9 @@ exports.run = async function({dir = '.', cover, watch, skipBuild}) {
   await runTests();
 
   if (watch) {
-    compiler.on('done', runTests);
+    compiler.on('done', () => {
+      runTests().catch(() => {}); // ignore error, since test output is already actionable enough
```
davidkearns: Should we log something here?

lhorie: I think logging would make the test output unnecessarily noisy. The stack trace is already part of the test report

lhorie:
!merge

PR #65, opened 2017-12-11
Build-time __dirname in browser test builds

rtsao:
Tape needs `__dirname` to function properly

Fixes https://github.com/fusionjs/fusion-cli/issues/66


rtsao:
!merge

PR #67, opened 2017-12-12
Add jest as a test runner

### LINKED ISSUE #68: Support jest as a test runner

KevinGrandon:
Creates a fusion-cli command, `fusion test-app`, which wraps and runs jest.

Exports a test() function which gets passed in a assert library for backwards compatibility.

Fixes #68


build/jest-framework-setup.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-env node */
+import {configure} from 'enzyme';
```
mlmorg: What if people don't use enzyme in their tests?

build/jest-framework-shims.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-env node */
+global.requestAnimationFrame = callback => {
```
mlmorg: What is this for?

KevinGrandon: Required for running react tests in node - they ask you to bring your own polyfills.

build/jest-framework-setup.js
KevinGrandon: I don't think it's necessary, but I think we should probably enable by default? We can add a config to turn this off though.

build/test-app-runtime.js
```diff
@@ -0,0 +1,54 @@
...
+
+module.exports.TestAppRuntime = function({dir = '.', watch = false, match}) {
+  const state = {proc: null};
+
+  this.run = () => {
+    this.stop();
+
+    let command = require.resolve('jest-cli/bin/jest.js');
+    let args = [
+      // '--no-cache',
```
rtsao: I'm guessing this should be deleted?

build/jest-transformer.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-env node */
+
+const transformer = require('babel-jest').createTransformer({
+  presets: [
+    'babel-preset-flow',
+    'babel-preset-es2015',
```
rtsao: Is the es2015 preset required given that we require Node 8+?

build/test-app-runtime.js
KevinGrandon: We can, though it's really useful for debugging.

build/jest-transformer.js
KevinGrandon: I'll try getting rid of it. 👍

KevinGrandon: Turns out this is required for modules to work. Might be able to do this with a smaller plugin, but since it's only for tests, I don't think it's too bad.

KevinGrandon:
!merge

PR #70, opened 2017-12-12
Add babel-preset-es2015

### LINKED ISSUE #68: Support jest as a test runner

KevinGrandon:
Turns out this is actually required for transpiling import statements within the jest files themselves. This should only transpile test code, so I'm not too worried about the extra size here.

Fixes #68


KevinGrandon:
This is a follow-up to #67 that was supposed to be included in the original PR but I forgot to push 🤐

PR #71, opened 2017-12-13
Use shared babel-preset file for transform

### LINKED ISSUE #68: Support jest as a test runner

KevinGrandon:
This uses our shared babel preset for tests so we don't need to duplicate a list of presets.

Fixes #68


KevinGrandon:
!merge

PR #74, opened 2017-12-14
Support running test-app against both node and browser targets

### LINKED ISSUE #68: Support jest as a test runner

KevinGrandon:
This preserves the universal/node/browser convention that we had previously established.

Fixes #68

### DIFF
--- a/build/babel-preset.js
+++ b/build/babel-preset.js
@@ -35,7 +35,12 @@ module.exports = function buildPreset(
           useBuiltIns: true,
         },
       ],
-      [require.resolve('babel-plugin-transform-cup-globals'), {target: target}],
+      ...(opts.transformGlobals
+        ? [
+            require.resolve('babel-plugin-transform-cup-globals'),
+            {target: target},
+          ]
+        : []),
       ...(target === 'browser'
         ? [
             [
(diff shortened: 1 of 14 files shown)


build/babel-preset.js
```diff
@@ -35,7 +35,12 @@ module.exports = function buildPreset(
           useBuiltIns: true,
         },
       ],
-      [require.resolve('babel-plugin-transform-cup-globals'), {target: target}],
+      ...(opts.transformGlobals
+        ? [
+            require.resolve('babel-plugin-transform-cup-globals'),
```
KevinGrandon: Needed to disable this for jest so we could inject our own globals.

test/fixtures/test-jest-app/stats.js
```diff
@@ -0,0 +1,19 @@
...
+module.exports.incrementCount = function() {
+  let count = getCount();
+  count += 1;
+  fs.writeFileSync(statsFile, String(count));
+}
+
+module.exports.getCount = getCount;
+
+module.exports.resetCount = function() {
+  fs.writeFileSync(statsFile, '0');
```
rtsao: I think one reason for parsing stdout instead of writing to a file is if in the future, JSDom is replaced with headless chrome, we can't really write to the filesystem and we'd need to change how these tests work.

KevinGrandon:
!merge

PR #75, opened 2017-12-15
Add updateSnapshot CLI arg for updating snapshots

### LINKED ISSUE #68: Support jest as a test runner

KevinGrandon:
Adds a CLI arg, `--updateSnapshot` which will pass through to jest and update test snapshots.

Fixes #68


test/fixtures/snapshots/snapshot-no-match.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
mlmorg: Possible to not have these comments in here? Having some non-jest-specific api for snapshot testing will allow us to change the underlying implementation in the future, if we want to.

KevinGrandon: Looking into it, but I don't believe so. I bet we could codemod the snapshot engine though if needed.

KevinGrandon:
!merge

PR #76, opened 2017-12-15
Add optional .fusionrc.js with babel field for custom plugins/presets

rtsao:
Resolves https://github.com/fusionjs/fusion-cli/issues/77

### Why not `.babelrc`?

- We want `{babelrc: false}` so nested configuration isn't read (i.e. in `node_modules`). This behavior will probably be removed from babel 7, but in the meantime this is important.
- `.babelrc` isn't meant for extension, it's meant to be the full source of truth. But we don't want people to need to explicitly extend from our config (which would be hard in the first place since we dynamically create lots of babel configuration).

### DIFF
--- a/build/load-fusionrc.js
+++ b/build/load-fusionrc.js
@@ -1,0 +1,27 @@
+/* eslint-env node */
+
+const fs = require('fs');
+const path = require('path');
+
+module.exports = function validateConfig(dir) {
+  const configPath = path.join(dir, '.fusionrc.js');
+  let config;
+  if (fs.existsSync(configPath)) {
+    config = require(configPath);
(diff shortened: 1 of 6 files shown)


rtsao:
!merge

build/load-fusionrc.js
```diff
@@ -0,0 +1,27 @@
...
+  } else {
+    config = {};
+  }
+  return config;
+};
+
+function isValid(config) {
+  return (
+    typeof config === 'object' &&
+    Object.keys(config).every(el => ['babel'].includes(el)) &&
```
angus-c: will throw if config is null

PR #81, opened 2017-12-18
Add enzyme snapshot serializer

### LINKED ISSUE #68: Support jest as a test runner

KevinGrandon:
This will allow for snapshot serialization with shallow wrappers by default. The library falls back to a solution which also works with React wrappers, but IMO we should encourage people to use shallow rendering where possible.

The test cases become quite elegant:
```js
test('Test case', assert => {
  const wrapper = shallow(<div />);
  assert.matchSnapshot(wrapper);
});
```

Fixes #68

### DIFF
--- a/build/jest-config.js
+++ b/build/jest-config.js
@@ -2,15 +2,16 @@
 
 module.exports = {
   cache: true,
-  rootDir: process.cwd(),
   globals: {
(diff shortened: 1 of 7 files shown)


KevinGrandon:
!merge

PR #89, opened 2017-12-19
Ensure correct CI test exit status code

rtsao:
Removing coverage for now (running out of memory on buildkite boxes)


.buildkite/nodeTests
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-./node_modules/.bin/nyc --reporter=text --reporter=json node test/index.js
+./node_modules/.bin/nyc --reporter=text --reporter=json node dist-tests/node.js || exit 1
```
KevinGrandon: This should probably be `test/index.js` still, right?

PR #92, opened 2017-12-23
Add __DEV__ to jest environment

KevinGrandon:
Fixes #90

Issue #91, opened 2017-12-23
react-native support for test-app


KevinGrandon:
Support running tests with react-native.

Possible fixes include exposing a moduleNameMapper configuration, or using babel config from .fusionrc.js.

We might also need to include the react-native serializer.

https://github.com/necolas/react-native-web/blob/master/docs/guides/getting-started.md#testing-with-jest

KevinGrandon:
For now we will just require people to do this in code.

PR #93, opened 2017-12-23
Use .fusionrc babel plugins

KevinGrandon:
Refs #91

(Currently a stacked diff off #92)


KevinGrandon:
Only change was rebasing the stacked diff. Landing.

PR #95, opened 2017-12-30
Move enzyme to peer dependencies

KevinGrandon:
Loading different versions of enzyme can cause issues running tests, so enforce only having a single version with peerDependencies.

Refs #68


KevinGrandon:
!merge

PR #101, opened 2018-01-03
Add support for cdnUrl and code coverage reporting

### LINKED ISSUE #100: Add support for cdnUrl and report code coverage

ganemone:
Fixes #100

### DIFF
--- a/build/test-runtime.js
+++ b/build/test-runtime.js
@@ -17,11 +17,8 @@ module.exports.TestRuntime = function({
     const base = '.fusion/dist/test';
     const server = path.resolve(dir, base, 'server/server-main.js');
     const client = path.resolve(dir, base, 'client/client-main.js');
-    // let command = require.resolve('unitest/bin/cli.js');
-    let command = 'unitest';
-    // let command = 'node';
+    let command = require.resolve('unitest/bin/cli.js');
     let args = [`--browser=${client}`, `--node=${server}`];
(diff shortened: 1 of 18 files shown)


build/test-runtime.js
```diff
@@ -17,8 +17,11 @@ module.exports.TestRuntime = function({
     const base = '.fusion/dist/test';
     const server = path.resolve(dir, base, 'server/server-main.js');
     const client = path.resolve(dir, base, 'client/client-main.js');
-    let command = require.resolve('unitest/bin/cli.js');
+    // let command = require.resolve('unitest/bin/cli.js');
```
KevinGrandon: Is it intentional to leave this commented?

test/index.js
```diff
@@ -2,8 +2,9 @@
 require('./compiler/api');
 require('./compiler/errors');
 require('./cli/dev');
-require('./cli/test');
+// require('./cli/test');
```
KevinGrandon: Should this be uncommented?

build/test-runtime.js
ganemone: good catch - updating

test/index.js
ganemone: this should be just deleted

KevinGrandon:
Feels like we might possibly be introducing false-positives here with the run-command, given that tests are failing in master. Looking into this more.

PR #107, opened 2018-01-04
Add browser reporting for server startup errors

KevinGrandon:
Currently the browser will spin forever and fail to respond while an error is shown in the console. This adds a top-level page which will display on server error.

image

Refs #108


build/server-error.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-env node */
+const React = require('react');
+const RedBox = require('redbox-react').RedBoxError;
+const ReactDOMServer = require('react-dom/server');
+
+function renderError(res, error) {
+  res.write(
+    '<!DOCTYPE html><html><head><title>Server render error</title></head><body>'
```
ganemone: Technically I think this is not a server render error, but a server startup error

KevinGrandon:
Current test errors are in master and un-related. Going to land this while we investigate the test errors in a separate PR.

Issue #108, opened 2018-01-04
Error reporting in dev


KevinGrandon:
Let's surface errors for faster dev cycles.

- [ ] Make sure sourcemaps work
- [x] Server startup errors
- [ ] Server render errors
- [x] Compilation errors
- [ ] Reload when server startup errors fixed
- [ ] Reload when server render errors fixed
- [ ] Reload when compilation errors fixed
- [ ] Trigger page updates for server startup errors
- [ ] Trigger page updates for server render errors
- [ ] Trigger page updates for compilation errors

mlmorg:
What is the status of this?

PR #110, opened 2018-01-05
Display SSR errors in the browser in dev mode

KevinGrandon:
Currently we just display internal server error. Rendering the error in the browser should lead to a better developer UX. Error screen looks like:

image

Refs #108


.buildkite/pipeline.yml
```diff
@@ -20,3 +20,8 @@ steps:
     plugins:
       'docker-compose#v1.7.0':
         run: fusion-cli
+  - name: ':jest:'
```
KevinGrandon: For some reason the same tests have isolation concerns under tape, but work fine under jest. They're easy to move around later once we figure out the isolation problems.

KevinGrandon:
Thanks for fixing package.json.

!merge

KevinGrandon:
!merge

PR #122, opened 2018-01-09
Merge jsdom and node coverage results

### LINKED ISSUE #121: Combine coverage reports
Right now we run jest against multiple environments. We should combine coverage information between the environments to create a single test coverage file.

KevinGrandon:
This creates multiple coverage folders depending on what test suites are run (node, jsdom), and merges them after coverage is collected into a single coverage/ folder. There is some weirdness around the jest process exiting before coverage is done writing to disk, so we wait until expected coverage files are present before merging.

Fixes #121

### DIFF
--- a/build/merge-coverage.js
+++ b/build/merge-coverage.js
@@ -1,0 +1,46 @@
+/* eslint-env node */
+const fs = require('fs');
+
+// Jest writes coverage after the process exits
+// so we need to poll and wait for all coverage files.
+const waitForAllCoverage = config => {
+  return Promise.all(
+    config.environments.map(
+      env =>
+        new Promise(resolve => {
+          const interval = setInterval(() => {
+            const coverageExists = fs.existsSync(
+              `${config.dir}/coverage-${env}/coverage-final.json`
+            );
+            if (coverageExists) {
+              resolve();
+              clearInterval(interval);
+            }
+          }, 100);
+        })
+    )
+  );
+};
+
+module.exports = function(config) {
+  return waitForAllCoverage(config).then(() => {
+    const createReporter = require('istanbul-api').createReporter;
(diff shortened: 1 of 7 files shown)


build/merge-coverage.js
```diff
@@ -0,0 +1,46 @@
...
+
+// Jest writes coverage after the process exits
+// so we need to poll and wait for all coverage files.
+const waitForAllCoverage = config => {
+  return Promise.all(
+    config.environments.map(
+      env =>
+        new Promise(resolve => {
+          const interval = setInterval(() => {
+            const coverageExists = fs.existsSync(
```
mlmorg: I thought exists was being deprecated?

build/merge-coverage.js
```diff
@@ -0,0 +1,46 @@
+/* eslint-env node */
+const fs = require('fs');
+
+// Jest writes coverage after the process exits
+// so we need to poll and wait for all coverage files.
```
mlmorg: What if there is an error and coverage is never written?

build/merge-coverage.js
KevinGrandon: It looks like exists() is being deprecated, but existsSync() is not.

build/merge-coverage.js
KevinGrandon: The node process should just exit in this case, killing the CLI as well. I'll go ahead and include a timeout here though in case coverage never gets written for some reason.

test/fixtures/test-jest-app/.gitignore
```diff
@@ -1 +1 @@
-coverage/
+coverage*/
```
mlmorg: We should make sure to make this change in the scaffold.

KevinGrandon:
!merge

PR #124, opened 2018-01-09
Fix issue with serving assets from memory

### LINKED ISSUE #123: Fix issue with serving assets from memory

ganemone:
Fixes #123

### DIFF
--- a/test/cli/dev.js
+++ b/test/cli/dev.js
@@ -5,6 +5,7 @@ const path = require('path');
 const test = require('tape');
 const {dev} = require('../run-command');
 const {promisify} = require('util');
+const request = require('request-promise');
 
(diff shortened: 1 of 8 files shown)


test/cli/dev.js
```diff
@@ -21,23 +22,37 @@ test('`fusion dev` works', async t => {
   t.end();
 });
 
-test('`fusion dev` works with assets', async t => {
+test.only('`fusion dev` works with assets', async t => {
```
mlmorg: Remove .only

ganemone: good catch!

KevinGrandon: (We should implement some sort of linter against this probably)

ganemone:
!merge

Issue #125, opened 2018-01-09
Incorporate the ability to supply a route prefix


AlexMSmithCA:
See proposed [RFC](https://github.com/fusionjs/rfcs/pull/2).

PR #126, opened 2018-01-09
Add support for `ROUTE_PREFIX` and associated tests

AlexMSmithCA:
#125

### DIFF
--- a/build/compiler.js
+++ b/build/compiler.js
@@ -88,14 +88,12 @@ function getConfig({target, env, dir, watch, cover}) {
   const destination = path.resolve(dir, `.fusion/dist/${env}/${side}`);
   const evergreen = false;
   const possibleESVersions = ['es5'];
-  const serverEntry =
-    env === 'test'
-      ? serverTestEntry
-      : path.join(__dirname, `../entries/server-entry.js`);
-  const clientEntry =
-    env === 'test'
-      ? browserTestEntry
-      : path.join(__dirname, `../entries/client-entry.js`);
+  const serverEntry = env === 'test'
+    ? serverTestEntry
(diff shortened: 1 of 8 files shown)


build/compiler.js
```diff
@@ -608,8 +603,7 @@ function Compiler({
       },
       reporter: null,
       serverSideRender: true,
-      // TODO(#14): handle route prefix here??
-      publicPath: '/_static/',
+      publicPath: '/_static', // keep in sync with defaults in environment-variables-plugin.js
```
AlexMSmithCA: Only relevant change in this file.  Remainder is `prettier`'s doing.

plugins/route-prefix-context-plugin.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-env node */
+
+const envVarsPlugin = require('./environment-variables-plugin');
+
+module.exports = function() {
+  const envVars = envVarsPlugin().of();
+  return function middleware(ctx, next) {
```
AlexMSmithCA: Goal is to strip the prefix from `ctx.url` and supply the prefix in `ctx.prefix` for any consumers that might rely on the prefix.

AlexMSmithCA:
I wanted to keep this backwards compatible for existing apps.  Note, however, that if a `ROUTE_PREFIX` is supplied certain plugins may not yet act as intended as they make hard-coded or implicit assumptions on route prefixes ('') and static asset ('/_static') paths.

build/compiler.js
KevinGrandon: Weird, linters are failing. Do you have an older prettier version installed?

dennisgl:
Looks good other than linting

PR #130, opened 2018-01-11
Add defer option to koa-static

### LINKED ISSUE #129: Add defer option to koa-static
Adding the `defer` option to the `koa-static` middleware tells it to `await next` before
setting the response. This allows composition with user middleware on the `/_static` route
which can be useful for things such as setting headers.

ganemone:
Fixes #129

### DIFF
--- a/bin/cli-runner.js
+++ b/bin/cli-runner.js
@@ -16,10 +16,6 @@ module.exports.run = args => {
     .version()
     .help();
   const cmdPath = path.resolve(cmdsDir, `${instance.argv._}.js`);
-  // we should always have process.cwd() be equal to the dir of the project being run
-  if (instance.argv.dir) {
-    process.chdir(instance.argv.dir);
-  }
   if (!fs.existsSync(cmdPath)) {
     const error = new Error(`'${instance.argv._}' is an invalid command\n`);
     // TODO(#3) ideally, the string from yargs.showHelp() should be part of the error message above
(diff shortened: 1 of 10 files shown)


mlmorg:
Do we have docs to show this composition?

ganemone:
I will add tests.

Regarding docs, we don't have anything for this specifically, but it should compose the same way all koa middleware can compose

mlmorg:
Maybe we should have docs for setting headers on static assets?

ganemone:
| Maybe we should have docs for setting headers on static assets?

Yea, I wanted to talk with everyone about a few things related to this. What prompted me to add this was issues with missing `cors` headers from our cdn. The way our cdn's are set up (afaik) they mimic headers they receive from the origin, which in our case is the fusion server. If this is a common enough use case, it might make sense to simply have cors be configured directly by fusion.

bin/cli-runner.js
```diff
@@ -16,6 +16,10 @@ module.exports.run = args => {
     .version()
     .help();
   const cmdPath = path.resolve(cmdsDir, `${instance.argv._}.js`);
+  // we should always have process.cwd() be equal to the dir of the project being run
```
rtsao: I'm not so sure about this. I think `process.cwd()` should only used as part of a default for `--dir`. Otherwise, I think `process.cwd()` should never be referenced anywhere in the CLI.

ganemone: I think that it helps a lot when dealing with using the file system. I don't see much of a downside to doing it, and the upside of making filesystem concerns consistent. What are the downsides that you see?

ganemone:
!merge

PR #140, opened 2018-01-13
Enable coverage for CI tests

KevinGrandon:


KevinGrandon:
This is outdated and wasn't working, so closing for now. Will look to revisit this some day, but I think that co-located tests and also moving to jest would help a lot here.

PR #156, opened 2018-01-20
Add --testFolder CLI argument for filtering test files

KevinGrandon:
This adds a new env argument for the CLI which will filter test files by folder. Example usage: `yarn test-app --testFolder=integration`. We currently do not support full globbing of filenames because that could potentially break the browser/node opinions of this CLI. Further globbing customization can be added as people need to support it in their apps.


KevinGrandon:
!merge

PR #168, opened 2018-01-24
Add support for routePrefix

### LINKED ISSUE #125: Incorporate the ability to supply a route prefix

ganemone:
Fixes #125

### DIFF
--- a/plugins/environment-variables-plugin.js
+++ b/plugins/environment-variables-plugin.js
@@ -13,9 +13,14 @@ if (!(env === 'development' || env === 'production')) {
   throw new Error(`Invalid NODE_ENV loaded: ${env}.`);
 }
 const prefix = load('ROUTE_PREFIX', '');
-const assetPath = '/_static';
+assert(!prefix.endsWith('/'), 'ROUTE_PREFIX must not end with /');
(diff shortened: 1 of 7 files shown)


plugins/environment-variables-plugin.js
```diff
@@ -13,9 +13,14 @@ if (!(env === 'development' || env === 'production')) {
   throw new Error(`Invalid NODE_ENV loaded: ${env}.`);
 }
 const prefix = load('ROUTE_PREFIX', '');
-const assetPath = '/_static';
+assert(!prefix.endsWith('/'), 'ROUTE_PREFIX must not end with /');
+const assetPath = load('FRAMEWORK_STATIC_ASSET_PATH', `${prefix}/_static`);
```
AlexMSmithCA: As an aside, we now prefix the `assetPath` with the `routePrefix` as opposed to concatenating them when necessary.  I don't think we ever want the asset path sans prefix so this should be fine, but wanted to point it out.

AlexMSmithCA:
Worth noting that we will likely want to register the env var-provided `ROUTE_PREFIX` into the DI system for user-land plugins to consume as well (e.g. react router; font plugin; cookie handlers; etc.)

ganemone:
!merge

PR #192, opened 2018-02-05
Fix issue with handling multiple errors in a row in dev

### LINKED ISSUE #191: Fix issue with handling multiple errors in a row in dev
There is a bug where the dev command can only gracefully handle one
error before crashing. We need to reset the listening state after an
error occurs.

ganemone:
Fixes #191


ganemone:
!merge

PR #199, opened 2018-02-06
Use jest projects

KevinGrandon:
Uses the jest project runner instead of our multi-spawn thing. This makes a lot of things easier to deal with, and should result in cleaner test output.

Passing multi-project tests:
image

Failing multi-project tests:
image

Passing single project tests:
image

Fixes #199


ganemone:
I think this still needs some work - turning caching on in example-trips-viewer causes seg faults. I think we can solve this with a custom cache key function, I'm testing that out now

KevinGrandon:
> I think this still needs some work - turning caching on in example-trips-viewer causes seg faults. I think we can solve this with a custom cache key function, I'm testing that out now

Interesting. I imagine this is due to config updates /w the cache? Manually clearing the cache probably makes it go away? Just wondering if this would persist on a typical yarn install.

KevinGrandon:
Just verified that I'm not getting segfaults when doing a CP of fusion-cli, so maybe it's possible you have a bad config saved?

ganemone:
!merge

PR #204, opened 2018-02-07
Add logging to debug intermittent test

KevinGrandon:


KevinGrandon:
Going to just land this PR to add logging for now to debug this intermittent test as it appears to not do it when you observe it of course.

PR #228, opened 2018-02-14
Fix assetUrl when used outside of app lifecycle

KevinGrandon:
Previously we were creating invalid assetUrls because `__webpack_public_path__` was not set yet. Work around this by reordering the the shared entry to be after the public path assignment.

Fixes #228

### DIFF
--- a/entries/server-entry.js
+++ b/entries/server-entry.js
@@ -29,7 +29,8 @@ into `__webpack_require__.p = ...` and uses it for HMR manifest requests
 // eslint-disable-next-line
 __webpack_public_path__ = getCompilationMetaData().webpackPublicPath + '/';
 
-// The shared entry must be imported after setting __webpack_public_path__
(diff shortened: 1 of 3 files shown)


entries/server-entry.js
```diff
@@ -31,6 +29,10 @@ into `__webpack_require__.p = ...` and uses it for HMR manifest requests
 // eslint-disable-next-line
 __webpack_public_path__ = getCompilationMetaData().webpackPublicPath + '/';
 
+// The shared entry must be imported after setting __webpack_public_path__
+// eslint-disable-next-line import/no-unresolved, import/no-extraneous-dependencies
+const main = require('__FRAMEWORK_SHARED_ENTRY__');
```
mlmorg: Can you add a comment as to why this is a require?

KevinGrandon: Sure, will expand on the above comment.

PR #238, opened 2018-02-17
Use the jest test runner by default

KevinGrandon:
* Migrate test-app -> test
* Proxy test-app commands -> test


KevinGrandon:
Oops, rebased this, but just going to land.

Issue #244, opened 2018-02-20
Bug with route prefix stripping


ganemone:
### Type of issue

Bug

### Description

There is a bug with route prefix stripping which sets `ctx.path` to `null` rather than `/` if the url matches the route prefix exactly.

### Current behavior

If the url matches the route prefix exactly, certain properties on ctx get incorrect values. For example, `ctx.path` gets set to null rather than '/'.

### Expected behavior

If the url matches the route prefix exactly, it should behave the same as an application receiving a request to '/'.

### Steps to reproduce

1. run a fusion application with the ROUTE_PREFIX environment variable set
2. make a request to `localhost:3000/$ROUTE_PREFIX`

PR #245, opened 2018-02-20
Fix issues with route prefix stripping 

ganemone:
#244


test/cli/build.js
```diff
@@ -138,6 +138,28 @@ test('`fusion build` works in production with default asset path and supplied RO
...
+  const {proc, port} = await start(`--dir=${dir}`, {
+    env: Object.assign({}, process.env, {ROUTE_PREFIX: '/test-prefix'}),
+  });
+  const rootRes = await request(`http://localhost:${port}/`);
+  t.equal(
+    rootRes,
+    'ROOT REQUEST',
+    'strips route prefix correctly for root requests'
+  );
+  const testRes = await request(`http://localhost:${port}/test`);
```
AlexMSmithCA: Shouldn't these have the route prefix in the request to ensure it gets stripped?

ganemone: yikes - I'm a fool

test/fixtures/prefix/src/main.js
```diff
@@ -0,0 +1,14 @@
+import App from 'fusion-core';
+
+export default async function() {
+  const app = new App('element', el => el);
+  app.middleware((ctx, next) => {
+    if (ctx.url === '/' && ctx.path === '/') {
+      ctx.body = 'ROOT REQUEST';
+    } else if (ctx.url === '/test' && ctx.path === '/test') {
```
AlexMSmithCA: We aren't stripping any route prefixes from `path`, right?

ganemone: It should strip it from any url related props on ctx, since I strip it before sending req/res to koa

PR #246, opened 2018-02-20
Upgrade and unpin dependencies

KevinGrandon:
It seems possible that this may be the cause of the scaffold failing due to loading multiple copies of jest because semvar ranges don't match. Another option might be to use peerDependencies, but I believe that this should fix this particular problem we're seeing.

Fixes #246


KevinGrandon:
!merge

PR #250, opened 2018-02-21
Fix route prefix stripping behavior

### LINKED ISSUE #244: Bug with route prefix stripping

ganemone:
Fixes #244

This was tested and verified as a prerelease. See https://github.com/fusionjs/fusion-cli/pull/245


ganemone:
!merge

PR #262, opened 2018-03-06
Add empty config for all node builtins with no default polyfill

### LINKED ISSUE #254: Improve experience of working with node builtins
### Type of issue

Feature request

### Description

Using node builtins currently requires employing some hacks due to the way we manage tree shaking in conjunction with our webpack node configuration. We should improve the experience of using node builtins.

### Current behavior

```js
// some-plugin.js
import fs from 'fs';
import {createPlugin} from 'fusion-core';

export default createPlugin({
  deps: {},
  provides: () => {
    // do some things with fs...
  }
}}

// src/main.js
import somePlugin from './some-plugin';
__NODE__ && app.register(somePlugin):
```

Running `dev` results in: 

```
error: (client) ./src/some-plugin.js
Module not found: Error: Can't resolve 'fs' in ...
```

This happens for two reasons.
1. we don't tree shake in development, so `src/some-plugin.js` is included in the client bundle
2. we set the `node.fs` webpack config to be `false`

### Expected behavior

The build should work, but the plugin won't be registered.

### Steps to reproduce

1. scaffold a fusion app
2. create a plugin that imports fs
3. apply that plugin in __NODE__ only
4. run `fusion dev`

### Proposed Solution

Long term, we can investigate ways of supporting tree shaking in development. Short term, we should set all `node` builtins to be `empty` (where possible) for the client build. This will allow the user to import them, but with throw errors if the user tries to use them. 

NOTE: Unfortunately, problems often arise when certain node builtins are set to 'empty', specifically surrounding testing, so we may need to make exceptions in certain cases.

ganemone:
Fixes #254

### DIFF
--- a/build/compiler.js
+++ b/build/compiler.js
@@ -198,30 +198,7 @@ function getConfig({target, env, dir, watch, cover}) {
       hints: false,
     },
     context: dir,
-    /**
-     * Don't polyfill or mock node things in the browser. Default settings can be found at:
-     * https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js
-     */
-    node: Object.assign(
-      {
-        // Polyfilling process involves lots of cruft. Better to explicitly inline env value statically
-        // Tape requires process to be defined
-        process: env === 'test' && target === 'web' ? 'mock' : false,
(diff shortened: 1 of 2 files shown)


ganemone:
!merge

PR #268, opened 2018-03-14
Add dynamic import test

KevinGrandon:
Currently there is no test case that exercises our dynamic import logic. This will ensure that we have some coverage for this functionality as we migrate to webpack 4.


KevinGrandon:
!merge

PR #277, opened 2018-03-15
Add named default async export test case

KevinGrandon:
Adds a test case that currently breaks:

image

To test this out in a CLI:
```
node /Users/keving/Uber/fusionjs/fusion-cli/bin/cli.js dev --port=63840 --no-open --dir=/Users/keving/Uber/fusionjs/fusion-cli/test/fixtures/named-async-main
```


KevinGrandon:
Rolling this into the new webpack4 branch.

PR #284, opened 2018-03-17
Remove react-hot-loader and add HMR test

KevinGrandon:
* Remove the React hot-loader babel plugin as we perform HMR with our own logic in client-entry.js.
* Adds integration tests for hot module replacement using puppeteer.
* Add two text fixture apps for testing HMR, one with the router, and one without.

Fixes #284


KevinGrandon:
!merge

PR #287, opened 2018-03-20
Implement pure annotation for createPlugin

KevinGrandon:
* Cherry-pick be9b2bf73c2b1b31cd9bb07a496aba9403f1ac11 in and update tests for babel7 transpilation changes.
* Adds a tree-shaking integration test in addition to babel plugin.

Fixes #287

### DIFF
--- a/test/fixtures/tree-shaking/src/main.js
+++ b/test/fixtures/tree-shaking/src/main.js
@@ -14,6 +14,8 @@ import defaultExportBrowserPlugin from './plugins/default-export-browser.js';
 import defaultExportNodePlugin from './plugins/default-export-node.js';
 import instrumentedAsPureBrowserPlugin from './plugins/instrumented-as-pure-browser.js';
 import instrumentedAsPureNodePlugin from './plugins/instrumented-as-pure-node.js';
+import {plugin as namedExportBrowserPlugin} from './plugins/named-export-browser.js';
+import {plugin as namedExportNodePlugin} from './plugins/named-export-node.js';
 
 export default async function() {
   const app = new App('element', el => el);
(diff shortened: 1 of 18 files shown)


test/fixtures/tree-shaking/src/main.js
```diff
@@ -10,8 +10,17 @@ import readline from 'readline';
...
 import tls from 'tls';
 
+import defaultExportBrowserPlugin from './plugins/default-export-browser.js';
+import defaultExportNodePlugin from './plugins/default-export-node.js';
+import instrumentedAsPureBrowserPlugin from './plugins/instrumented-as-pure-browser.js';
+import instrumentedAsPureNodePlugin from './plugins/instrumented-as-pure-node.js';
+
 export default async function() {
   const app = new App('element', el => el);
+  __BROWSER__ && app.register(defaultExportBrowserPlugin);
```
ganemone: I think it would be really useful to test named exports also.

test/fixtures/tree-shaking/src/plugins/instrumented-as-pure-node.js
```diff
@@ -0,0 +1,8 @@
+import {createPlugin} from 'fusion-core';
+
+const PLUGIN = /*@__PURE__*/ createPlugin({
```
ganemone: What is the purpose of the instrumented as pure files?

test/fixtures/tree-shaking/src/main.js
KevinGrandon: Good idea, test added.

test/fixtures/tree-shaking/src/plugins/instrumented-as-pure-node.js
KevinGrandon: Just thought it would be good to have a test case for this, in case that folks want to instrument their own code.

PR #290, opened 2018-03-22
Fix bundle splitting in production builds

KevinGrandon:
This fixes our production building in a way that continues to use the ChunkManifestPlugin. I'm not sure whether or not this is actually needed.

Fixes #290


KevinGrandon:
!merge

PR #291, opened 2018-03-22
Release v1.2.5

KevinGrandon:


AlexMSmithCA:
Is the only difference the version bump? :O

KevinGrandon:
> Is the only difference the version bump? :O

Ah, nah, I think it's because I opened before landing the other PR. This fixes the production build issues. We should fix the bot to update the link probably.

KevinGrandon:
!merge

PR #293, opened 2018-03-22
Set cache control headers on asset requests

### LINKED ISSUE #292: Set cache control headers on asset requests

ganemone:
Fixes #292


ganemone:
!merge

mlmorg:
Would this get set on assets served locally (e.g. non-hashed development files)?

ganemone:
No - we serve those via webpack-dev-middleware so we can serve directly from memory

PR #296, opened 2018-03-25
Add test case with failing production build

KevinGrandon:
Slowly built up this example to emulate a failing build since introducing webpack 4.

**Steps to reproduce**

```
cd test/fixtures/dynamic-import-app/
node ../../../bin/cli.js build --production
```

**Expected results**

The app will build properly

**Actual results**

The app fails to build with an error like:
```
error: (server) /.../fusion-cli/build/chunk-id-loader.js!./src/split-route.js
Module build failed: Error: no result from manifest for /.../fusion-cli/test/fixtures/dynamic-import-app/src/split-route.js
    at _generateSource (/.../fusion-cli/build/chunk-id-loader.js:27:24)
    at <anonymous>
 @ ./src/split-route.js 16:30-158
 @ ./src/routes.js
 @ ./src/root.js
 @ ./src/main.js
 @ /.../fusion-cli/entries/server-entry.js
 @ multi /.../fusion-cli/entries/server-entry.js

```


build/chunk-module-manifest-plugin.js
```diff
@@ -22,8 +22,18 @@ class ChunkModuleManifestPlugin {
           const chunkIdsByFile = new Map();
           chunks.forEach(c => {
             const chunkId = c.id;
-            const files = Array.from(c.modulesIterable, m => m.resource);
-            files.forEach(path => {
+
+            let additionalFileDeps = [];
+            const files = Array.from(c.modulesIterable, m => {
+              if (m.rootModule && m.rootModule.resource) {
+                additionalFileDeps = additionalFileDeps.concat(
```
ganemone: I'm not sure this is the correct fix. It might be, but I was able to fix this issue with a different set of changes. Let me put up my PR and we can compare.

KevinGrandon: I think both will probably work, but yours is cleaner, thanks!

KevinGrandon:
Better implementation of fix landed in #298.

Issue #182, opened 2018-02-02
Transpile node_modules


rtsao:
Currently, node_modules are not transpiled.

This means when a package exported as ES2015+ is consumed, the ES5 target browser bundle will contain ES2015+ code, which is unexpected. Instead, we should transpile node_modules in browser bundles.

PR #300, opened 2018-03-27
Add transpilation of node_modules

rtsao:
Fixes https://github.com/fusionjs/fusion-cli/issues/182

### DIFF
--- a/build/babel-fusion-preset.js
+++ b/build/babel-fusion-preset.js
@@ -1,0 +1,30 @@
+/* eslint-env node */
+
+/**
+ * This babel config is used for fusion-related compilation,
+ * including non-standard language features (e.g. React and Flow).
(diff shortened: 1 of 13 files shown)


rtsao:
It will be slower, but it's unclear by how much (I haven't done any profiling). Anecdotally it seemed only slightly slower. The extra overhead will per-file, so incremental rebuilds will mitigate the cost of this change at least.

rtsao:
Oh, turns out an babel config overrides feature was added in Babel 7, so this can probably be done without using another webpack loader. This would eliminate the performance overhead of the extra loader in the current implementation. Let me try refactoring to use that instead.

rtsao:
No, only ES2017+ stuff is transpiled globally.

rtsao:
!merge

PR #301, opened 2018-03-27
Fix redirects

ganemone:
There was an issue with redirects where the 
dev proxy followed the redirect, then proxied
the response back to the browser. This caused 
the browser url to not be updated. This PR updates
the dev proxy to not follow redirects, instead proxying
the redirect to the browser.


ganemone:
!merge

PR #302, opened 2018-03-28
Add test case for simulation race condition

KevinGrandon:
**Description**
There's currently a bug when multiple simulation tests are run in parallel. I believe that this case reproduces the issue as expected.

**Steps to reproduce:**
```bash
test/fixtures/test-jest-react-app
CI=true node ../../../bin/cli.js test --configPath=../../../build/jest-config.js
```

**Error stack:**
```
  ● node simulate

    TypeError: originalMiddleware is not a function

      38 |       try {
      39 |         // $FlowFixMe
    > 40 |         return fn(context, function next() {
      41 |           return dispatch(i + 1);
      42 |         });
      43 |       } catch (err) {
      
      at ../../../node_modules/fusion-react/src/plugin.js:27:18
      at dispatch (../../../../fusion-core/dist/index.js:40:16)
      at next (../../../../fusion-core/dist/index.js:41:18)
      at jwtMiddleware (../../../node_modules/fusion-plugin-jwt/src/jwt-server.js:112:15)

```

The fusion-react package bump here fixes the issue.

Fixes #302


test/cli/test.js
```diff
@@ -336,3 +336,13 @@ test('`fusion test --debug --env=jsdom,node`', async t => {
 
   child.kill();
 });
+
+test.only('`fusion test` simultaneous simulation tests across envs', async t => {
```
ganemone: test.only

KevinGrandon:
I think that we covered this ok actually in the test that we did for fusion-react, and there is a better test-case we can do for the real issue here. PR incoming.

PR #307, opened 2018-03-28
Fix bug with getting chunk ids

ganemone:


ganemone:
Planning on writing a test for this

Issue #314, opened 2018-04-03
production builds using fusion-cli@1.3.0-1 fail on ios Safari/UIWebView


koulmomo:
### Type of issue
:bug:

### Description
I have only been able to repro this on an actual physical iphone, iOS simulator does **NOT** error.

error in console:
```
Error: Loading chunk 3 failed.
(error: production_url_for_fusion_app/_static/3-8d3fd1c14cbf6fa2672f.js)
```

Error occurs in a file that looks like:
`production_url_for_fusion_app/_static/client-vendor-3249340eddecd23065de.js`

I have confirmed that the error DOES NOT occur with version `1.1.0` of fusion-cli.

### Current behavior

Client side error occurs in mobile safari + UIWebView causing the page to not finish client-side hydrating.

### Expected behavior

It should not do that

### Your environment

* fusion-cli version: 1.3.0-1

* Node.js version (`node --version`): 8.9.3

* npm version (`npm --version`): 5.6.0

* Operating System: iphone 7+, iOS 11.2.6

KevinGrandon:
Thank you for the detailed bug report and sorry you're running into issues.

> Client side error occurs in mobile safari + UIWebView

Does this mean both the mobile safari browser, as well as a custom in-app page using UIWebView?

I'm wondering if the bug reproduces with WKWebView as well.

rtsao:
There's an outstanding bug with Safari where it doesn't send cookies for script tags with `crossorigin="anonymous"` attributes, even when the source url is same-origin. See:
 - https://bugs.webkit.org/show_bug.cgi?id=171566
 - https://bugs.webkit.org/show_bug.cgi?id=171550

As a result of this, any chunks loaded by webpack will fail in Safari if they require cookies to be sent (because webpack uses `crossorigin="anonymous"`).

rtsao:
Still some issues from this...

PR #324, opened 2018-04-05
Prevent crossorigin attributes for dynamically loaded scripts when same-origin

rtsao:
Fixes https://github.com/fusionjs/fusion-cli/issues/314


rtsao:
!merge

PR #334, opened 2018-04-11
Fix crossorigin attribute logic for Safari

rtsao:
Turns out `script.src` and `script.getAttribute('src')` yield different results for relative sources, which made the logic faulty. `script.src` yields an absolute URL whereas `script.getAttribute('src')` yields the relative URL.

Fixes https://github.com/fusionjs/fusion-cli/issues/314

The existing test had a false negative because tests are served via `http://` instead of `https://`. The new implementation avoids this pitfall.

### DIFF
--- a/build/crossorigin-auto-attribute-plugin.js
+++ b/build/crossorigin-auto-attribute-plugin.js
@@ -12,11 +12,11 @@
  * As a result of this, any chunks loaded by webpack will fail in Safari
  * if they require cookies to be sent (because webpack uses crossorigin="anonymous").
  *
- * This plugin ensures that the crossorigin attribute is omitted
- * unless the script src begins with `https://`.
+ * This plugin ensures that the crossorigin attribute is omitted if the script
+ * src matches the page origin.
  *
- * Note: if for some reason an absolute path is used even for same-origin scripts,
- * cookies may be erronously not be sent in Safari.
+ * Note: `script.src` accesses the `src` property
+ * (as opposed to attribute) which yields an absolute URL per HTML5 spec
  */
 
 class CrossoriginAutoAttributePlugin {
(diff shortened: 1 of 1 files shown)


build/crossorigin-auto-attribute-plugin.js
```diff
@@ -29,7 +29,7 @@ class CrossoriginAutoAttributePlugin {
         source => {
           return source.replace(
             /script\.src = [^;]*;/,
-            '$& if (!script.src.match(/^https:\\/\\//)) {script.crossOrigin = void 0;}'
+            '$& if (!script.src.indexOf(window.location.origin)) {script.crossOrigin = void 0;}'
```
KevinGrandon: `indexOf()` will return a 0 if it starts at the beginning, so should we compare with -1 instead?

rtsao: That's actually how it's intended to work.

When expanded, the logic looks something like this:

```js
var isSameOrigin = script.src.startsWith(window.location.origin);
// equivalent to:
var isSameOrigin = script.src.indexOf(window.location.origin) === 0;
// equivalent to:
var isSameOrigin = !(script.src.indexOf(window.location.origin) !== 0);
// equivalent to:
var isSameOrigin = !(Boolean(script.src.indexOf(window.location.origin)));
// equivalent to:
var isSameOrigin = !(script.src.indexOf(window.location.origin));

if (isSameOrigin) {
  removeCrossOriginAttribute();
}

```

I picked the smallest implementation (excluding `startsWith` because that would require polyfills) since it's going directly into the webpack bootstrap code and assigning variables, etc. is probably a bad idea. Also I don't think Uglify will be able to perform the simplification above.

lhorie: Maybe add a comment then?

rtsao: There's already a fairly extensive comment above that describes what this plugin does and why it is necessary. With that context, I don't think there should be any confusion.

Issue #274, opened 2018-03-15
Test coverage is not covering the intended files


ardok:
Test coverage is not covering all src files as intended. Right now, it only covers the tests that you include in the tests. Resulting in false coverage report.

### Type of issue

Bug

### Description

You just need to specify `collectCoverageFrom` in jest config. Something like this:
```
{
  "collectCoverageFrom": [
    "!**/node_modules/**",
    "<rootDir>/src/client/**/*.js",
    "<rootDir>/src/server/**/*.js",
    "<rootDir>/src/shared/**/*.js"
  ]
}
```

This is my setup in my project: https://gist.github.com/ardok/851685bf255fdd9533797055fbc355f3

### Current behavior

Only report coverage on the included files in the test.

### Expected behavior

Coverage report should cover the whole project.

### Steps to reproduce

`fusion test-app --coverage`

### Your environment

* fusion-cli version: 1.1.0

* Node.js version (`node --version`): 8.9.4

* npm version (`npm --version`): 5.6.0

* Operating System: Mac OS

KevinGrandon:
Great bug report, thank you for the code example!

ardok:
Np!
I just realized that on my project too. I was so happy I got to 48% line coverage until I realized that it's just what I've included in the test.
Looked around online and finally fixed it with some tweaks on the report coverage combination and jest collectCoverageFrom, and voila, 14% line coverage now 😄 😄 😄

PR #336, opened 2018-04-12
Ensure coverage reports include .js files that aren't imported in tests

rtsao:
Fixes https://github.com/fusionjs/fusion-cli/issues/274


test/fixtures/test-jest-app/bar.js
```diff
@@ -0,0 +1,3 @@
+export function bar() {
```
KevinGrandon: I might consider renaming this file to be specific to our test case.

rtsao:
!merge

rtsao:
!merge

Issue #259, opened 2018-02-26
fusion dev crashes when attempting to bundle a module that is not installed


ganemone:
### Type of issue

Bug

### Description

When running `fusion dev` the process crashes if you try to bundle a module that is not installed. We should try and handle this gracefully if possible.

### Current behavior

The process crashes with the following error:

```sh
Error: Cannot find module '<some module>'
    at Function.Module._resolveFilename (module.js:538:15)
```

### Expected behavior

We should attempt to handle this gracefully, and restart the compilation on a change to node_modules.

### Steps to reproduce

1. scaffold a fusion project
2. add an import statement for a module not installed
3. run `fusion dev`

PR #341, opened 2018-04-18
Fix crash on missing node_module

rtsao:
`resolve-from` throws an error by default if resolution fails. Instead, in this case, we should just leave the request as-is and let normal webpack error handling occur.

Fixes https://github.com/fusionjs/fusion-cli/issues/259


rtsao:
!merge

PR #350, opened 2018-04-26
Change test metadata disk logging to be based on env var

rtsao:
This is better than the existing behavior for a couple reasons:
- Logging based on CI env was somewhat arbitrary. It may not be desired in CI, and it may be desired outside of CI. Introducing a specific env var for this is more explicit
- Distinct test runs can output to separate files if a different filename env var is set


rtsao:
!merge

PR #354, opened 2018-05-08
Greenkeep build dependencies

KevinGrandon:
For big fixes and performance improvements. This should also get rid of a console deprecation warning when building fusion apps.


KevinGrandon:
!merge

PR #361, opened 2018-05-14
Add custom winston libdef

KevinGrandon:
This is the same libdef from fusion-plugin-universal-logger, but with one fix for the transports.

This is necessary otherwise we see flow errors such as:

image


KevinGrandon:
!merge

PR #366, opened 2018-05-21
Update winston flow type library

KevinGrandon:
To keep in lockstep with the change here: https://github.com/fusionjs/fusion-plugin-universal-logger/pull/109/files#diff-00e82586d3d52e3067526f8cf3bfe2d7R53


AlexMSmithCA:
Good catch!  Thanks for handling this

PR #369, opened 2018-05-24
Add support for checking against __NODE__ and __BROWSER__

ganemone:
It seems that even having the cup globals plugin before the tree-shake plugin does not guarantee that the `__NODE__` and `__BROWSER__` identifiers will be replaced with literals. This adds support for checking against the `__NODE__` and `__BROWSER__` literals.

### DIFF
--- a/build/babel-plugins/babel-plugin-transform-tree-shake/test/index.js
+++ b/build/babel-plugins/babel-plugin-transform-tree-shake/test/index.js
@@ -4,12 +4,13 @@ const fs = require('fs');
 const test = require('tape');
 const {transformFileSync} = require('@babel/core');
 const plugin = require('../');
+const {globalsPreset} = require('../../../babel-fusion-preset.js');
 
-test('boolean expression transformed', t => {
+test.only('boolean expression transformed', t => {
   const output = transformFileSync(
     __dirname + '/fixtures/input-boolean-expression',
     {
-      plugins: [plugin],
+      presets: [[globalsPreset, {target: 'browser', transformGlobals: true}]],
     }
   );
   const expected = fs
--- a/build/babel-plugins/babel-plugin-transform-tree-shake/test/index.js
+++ b/build/babel-plugins/babel-plugin-transform-tree-shake/test/index.js
@@ -6,7 +6,7 @@ const {transformFileSync} = require('@babel/core');
 const plugin = require('../');
 const {globalsPreset} = require('../../../babel-fusion-preset.js');
 
-test.only('boolean expression transformed', t => {
+test('boolean expression transformed', t => {
   const output = transformFileSync(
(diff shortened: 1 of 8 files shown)


build/babel-plugins/babel-plugin-transform-tree-shake/test/index.js
```diff
@@ -4,12 +4,13 @@ const fs = require('fs');
 const test = require('tape');
 const {transformFileSync} = require('@babel/core');
 const plugin = require('../');
+const {globalsPreset} = require('../../../babel-fusion-preset.js');
 
-test('boolean expression transformed', t => {
+test.only('boolean expression transformed', t => {
```
rtsao: 😉

ganemone: LOL

clean.sh
```diff
@@ -0,0 +1,3 @@
+for fixture in ./test/fixtures/*; do
+  rm -rf ./test/fixtures/$fixture/node_modules;
```
rtsao: I think something like this would be more robust:
```
git clean -xfd test/fixtures
```

rtsao: This also won't accidentally delete checked-in fixture `node_modules` (such as in the `transpile-node-modules` fixture).

ganemone: I think that is less robust, as it removes all files rather than just the ones we want to remove. Also, that would remove untracked files that someone might be working on. For example, if you added a new fixture and ran tests, those fixtures would be removed. We could potentially use the `-X` flag but I also don't want to worry about different versions of `git` on various machines, etc.

package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "lint": "eslint .",
-    "test": "node test/index.js"
+    "test": "./clean.sh && node test/index.js"
```
KevinGrandon: Can the test script clean up after itself instead? Maybe even just using fs?

clean.sh
rtsao: Good point about untracked files. Using`-X` definitely makes sense.

Is different git versions actually a concern? Even git `1.6.0`, (released in 2008) [supports all those flags](https://git-scm.com/docs/git-clean/1.6.0). Support probably goes even further back, so it doesn't seem like different git versions is a concern.

Also I think using git for this rather than shell scripts arguably makes it more portable (i.e. windows).

ganemone: Good point about portability, sometimes I forget windows exists lol. I will update to use the -X flag

package.json
ganemone: I find cleaning up before is more robust because in many cases random failures or early exits can put you in a bad state if you clean up after. Since these files are gitignored it doesn't matter if they exist when developing, we just need to be sure the space is clean before running tests.

PR #375, opened 2018-05-25
Handle else, preserve unreachable code

rtsao:
- Fixes alternate cases w/ `IfStatement`s
- Removes pruning of unreachable code, instead only tree shake unreachable imports.


rtsao:
!merge

PR #379, opened 2018-05-31
Fix coverage inclusion pattern

lassedamgaard:
This broke in 5253325dd529e2cd154f950b6a84883c6a004111 because the pattern does not recursively match js files.


KevinGrandon:
!merge

PR #380, opened 2018-05-31
Fix test for RedBoxError react type violation

blackswanny:
test `fusion dev` top-level error failed due to react type violation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ node_modules/
 .fusion/
 .nyc_output/
 yarn-error.log
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ node_modules/
 .fusion/
 .nyc_output/
 yarn-error.log
-
+.idea/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -3,4 +3,4 @@ node_modules/
 .fusion/
 .nyc_output/
 yarn-error.log
-
+.idea/
```
KevinGrandon: Not sure if we want to include ide-specific files in our gitignore files - might be better to add this to your user directory? I believe you can have your own files listed in a global `~/.gitignore` file which would work for all projects.

build/server-error.js
```diff
@@ -12,12 +12,21 @@ const React = require('react');
 const RedBox = require('redbox-react').RedBoxError;
 const ReactDOMServer = require('react-dom/server');
 
-function renderError(error /*: any */) {
+function renderError(error /*: any = {} */) {
```
KevinGrandon: What's the purpose for the `= {}` assignment here? If it's for a default assignment it seems like that should be outside of the comment?

.gitignore
blackswanny: removed

build/server-error.js
blackswanny: yes, default value.In this case the very last fallback for Object version of param will not faill if somebody tries to use undefined instead.
but outside of comment flow type check fails

PR #384, opened 2018-06-01
Change assert plugin path to relative

blackswanny:


plugins/assets-plugin.js
```diff
@@ -12,13 +12,15 @@ const {createPlugin, getEnv} = require('fusion-core');
 const path = require('path');
 const mount = require('koa-mount');
 const serve = require('koa-static');
+const parseUri = require('../lib/parseUri.js');
```
rtsao: I'm not sure I understand the purpose of adding this code.

Won't [`url.parse`](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) or the [`URL` constructor](https://nodejs.org/api/url.html#url_the_whatwg_url_api) suffice for this use case?

test/cli/build.js
```diff
@@ -93,7 +93,10 @@ test('`fusion build` works in production with a CDN_URL', async t => {
     'Server Entry file sourcemap gets compiled'
   );
   const {res, proc} = await start(`--dir=${dir}`, {
-    env: Object.assign({}, process.env, {CDN_URL: 'https://cdn.com/test'}),
+    env: Object.assign({}, process.env, {
```
KevinGrandon: Can we also make a new test case instead of updating this one? Would be nice to be sure that we have a test case for only the new functionality, while preserving coverage for older cases.

plugins/assets-plugin.js
blackswanny: URL is just a subset of URI for web. This is a bug about server-side issue.
So URI includes much more cases to check for relative path, including: other protocols (ftp, file, socket), backslash and slash platform-based cases, etc

KevinGrandon: Fairly sure we only need to handle HTTP and HTTPS for this use-case right now, and would be better to reduce amount of code if possible. We can re-evaluate if we ever need a different protocol.

Though I tested it out and it seems that `new URL` does seem to work with various protocols, FTP included.

blackswanny: I was trying to save balance between overcomplicating things and doing it right. This is a very small piece of code, less than 1Kb and it serves both needs, not only detects absolute path (all cases), but also extracts relative with no risk to fail. So if this code is not run too often (cause it's still slightly slower than just string manipulations), I would rather leave it.

test/cli/build.js
blackswanny: yes. added new one

plugins/assets-plugin.js
rtsao: The benefit of supporting non-HTTP(S) seems a bit unclear to me. Should we do that? Why not merely throw an error rather than try to support that use case (which seems somewhat dubious in the first place).

blackswanny: while discussion I understood that Path is correct, but due to specifics of mounting assets I just need to truncate relative part of it for this particular case. So it doesn't sound like an error so far. Otherwise, we need to decide, what to do with FRAMEWORK_STATIC_ASSET_PATH. Is it used somewhere else? Why is it absolute? Is it still needed? I am not sure I can answer it without help

KevinGrandon: I'm not sure if truncation is the correct solution here. Should we even be running this logic if FRAMEWORK_STATIC_ASSET_PATH is not a relative path?

blackswanny: I don't know, but FRAMEWORK_STATIC_ASSET_PATH is what comes here. And 100% if it is absolute, the `mount` function will raise exception and does not work

blackswanny:
The PR is no longer valid as we found another way to fix it

PR #385, opened 2018-06-04
Remove custom webpack chunk hash and external manifest plugins

rtsao:
The custom chunk hash plugin currently being used will erroneously yield identical hashes in some cases.

This PR makes the following changes
- Remove the custom chunk hash plugin (which should not be required in Webpack 4 anyways)
- Remove the custom external manifest plugin. This plugin is currently doesn't work correctly (as a result of Webpack 4), but even if it did, the manifest global variable isn't set anywhere. In any case, it seems using the new Webpack 4`optimization.runtimeChunk` option (and inlining this chunk) is preferred over external manifest plugins.

We should revisit optimization of bundle splitting/assets/chunking/naming configuration, but for now this PR resolves correctness problems. We will likely want to take a different approach in the future anyways (as part of per-browser bundles).


rtsao:
Yeah, I had trouble reproducing this issue constructively via a simple fixture. I think I could create one if I spent more time working backwards from a production app where I can reproduce this.

That said, bcause this PR just removes custom plugins, in theory we move closer to vanilla webpack and can rely more on webpack test coverage for things isolated to build-related concerns.

rtsao:
!merge

PR #392, opened 2018-06-14
Add assumeNoImportSideEffects option to .fusionrc.js

rtsao:
This option allows apps to opt into aggressive tree shaking / import pruning by assuming all import statements are side effect free

This means:
- Unused specifiers of import statements can be tree shaken (provided the source module is comprised of simple re-export statements)
- Completely unused import statements (all specifiers are unused) can be pruned entirely

We may eventually want to turn on some form of behavior on by default. Also, our custom babel tree shaking babel plugin is redundant once dead branches in LogicalExpressions removed in Webpack itself.


rtsao:
!merge

Issue #394, opened 2018-06-15
Support customizing coverage inclusion/exclusion


lassedamgaard:
### Type of issue
Feature request

### Description
I want to be able to exclude helper scripts outside of `/src` from coverage reporting.

### Current behavior
All JS files excluding `node_modules` and `__integration__` are included in coverage reporting. The only way to override this is to pass in a custom jest config which then (AFAICT) overrides the built-in configs entirely.

### Expected behavior
Either there should be an option to merge in jest config options instead of replacing the built-in config, or an dedicated option to control the cover inclusion (or exclusion) pattern.

### Your environment

* fusion-cli version: 1.6.3

* Node.js version (`node --version`): 8.11.2

rtsao:
Have you considered using [`istanbul ignore` comments](https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md#ignoring-code-for-coverage-purposes)?

We currently don't provide any special configuration into `babel-plugin-istanbul`, which means [it will look for a "nyc" field in your package.json](https://github.com/istanbuljs/babel-plugin-istanbul/blob/master/src/index.js#L36). This means [configuration exclusions in this manner](https://github.com/istanbuljs/nyc#configuring-nyc) should already work.

KevinGrandon:
I tested `/* istanbul ignore file */`, and it seemed to work great. This seems like the ideal short-term solution so we can maintain a smaller surface API for less things to go wrong. Happy to discuss further if you're not happy with adding that comment.

lassedamgaard:
istanbul comments work as expected but I think longterm this is not an ideal solution:

1. Usually you want to exclude whole directories or classes/patterns of files, not individual ones. For examples, fixtures, repo helper scripts, config files (e.g. `.someconfig.js`). Doing this manually is tedious and error prone and using the Jest coverage config you should be able to avoid it.

2. Fusion AFAIU aims to wrap or hide the underlying toolchain, for example not allowing you to configure Webpack and not officially supporting custom Babel config. It then to me seems inconsistent to expose the fact that code coverage is provided by istanbul (via Jest). There's nothing fundamentally wrong with doing this I guess, but it should then be documented which hooks into the build/test toolchain are supported.

KevinGrandon:
Good points, we'll continue to monitor this and collect feedback around usage.

> it should then be documented which hooks into the build/test toolchain are supported.

Definitely agree about the lacking documentation. We'll be sure to add this information to our docs.

PR #395, opened 2018-06-15
Add support for specifying additional coverage options

KevinGrandon:
This will allow engineers to pass in additional overrides or files to ignore.

Refs #394


KevinGrandon:
Let's go with `/* istanbul ignore file */`

PR #397, opened 2018-06-19
Bind the HTTP server to the HttpServerToken

derekjuber:
This re-organizes the logic in the server-entry so that the node server is available to be registered as a token. This will allow using it as a dependency within a plugin to add support for websockets.

This also replaces the current custom proxy with `http-proxy` that allows us to also proxy websocket connections.


.prettierrc
```diff
@@ -0,0 +1,13 @@
+{
```
KevinGrandon: None of our plugins have this, instead leveraging autofixing in eslint-config-fusion. I would prefer to isolate this change, and add this file everywhere if necessary, but I don't think it is?

KevinGrandon:
> Also adds a prettier configuration file as well as a prettier yarn script. The configuration was copied over from the one currently used internally in the linter configurations and rather than fix the errors by hand it was easier to leverage prettier.

`eslint . --fix` should fix all errors, did that not work?

derekjuber:
Sure, if that's our process I'll remove the prettier updates

entries/server-entry.js
```diff
@@ -64,17 +67,21 @@ export async function start({port, dir = '.'} /*: any */) {
   });
 
   return new Promise(resolve => {
-    server.listen(port, () => {
-      // eslint-disable-next-line no-console
-      resolve(server);
-    });
+    server &&
+      server.listen(port, () => {
+        // eslint-disable-next-line no-console
```
KevinGrandon: Just noticed that this eslint thing can be removed (not related to your change though).

AlexMSmithCA:
Could we add a regression test to ensure that `HttpServerToken` resolves correctly?

derekjuber:
@KevinGrandon: Anything else I need to do to get this landed?

derekjuber:
!merge

PR #399, opened 2018-06-21
Fix assetUrl and json files with webpack 4

KevinGrandon:
Webpack 4 started handling parsing of JSON files automatically and abstracted away the "type" of file. We can restore the webpack 3 behavior by including some loader config to just set the type of json files, which will allow json files to work inside of assetUrl again.

More context: https://github.com/webpack/webpack/issues/6586#issuecomment-368677035


KevinGrandon:
!merge

PR #401, opened 2018-06-22
Fix json importing

KevinGrandon:
We need to explicitly use the json-loader, or importing json content will not work due to our loader config of overriding the type. It seems that the JSON in webpack 4 won't work, but the json-loader package will. I haven't investigated why.


KevinGrandon:
!merge

PR #405, opened 2018-06-26
Prevent minification bug that causes const reassignment

lhorie:
Fix const reassignment minification bug (see https://github.com/mishoo/UglifyJS2/issues/2842 )


lhorie:
@rtsao yeah good call

PR #406, opened 2018-06-26
Refactor i18n key discovery

rtsao:
The main purpose of this refactor is to eliminate eager `require`ing of `import()`ed modules in the server bundle. To to this, this PR implements a few internal changes to i18n key discovery (which previously relied on this behavior):

**Summary of changes**
- Remove eager `require`ing from the custom webpack harmony import template
- Change i18n babel plugin implementation
   - *Old behavior*: inline i18n singleton instrumentation code into each file
   - *New behavior*: Extracting i18n key info from each file and make it available to webpack
- Add i18n singleton instrumentation into server entry (using loader that consumes extracted i18n key info)
- Add webpack plugin to invalidate extracted i18n key info when files are modified
- Removed xp babel plugin from webpack which isn't currently being used (and would need to be refactored anyways)


test/fixtures/split-translations/src/root.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+
+import React from 'react';
+import Router, {Route, Switch, Link} from 'fusion-plugin-react-router';
+import {Translate} from 'fusion-plugin-i18n-react';
+import {split} from 'fusion-react-async';
```
KevinGrandon: nit: seems like we don't use this import in this file. I'm surprised linters are passing, maybe our config is incorrect?

build/i18n-manifest-instrumentation-loader.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  if (entries.length) {
+    return `
+      var map = require('fusion-plugin-i18n/chunk-translation-map');
+      if (module.hot) {
+        // invalidate existing translations if hot reloading
+        map.translations = new Map();
+      }
+      [${entries.join(',')}].forEach(args => map.add(...args));
```
KevinGrandon: Isn't entries an array of strings here? Just wondering about the `[arr.join()].forEach` logic, something feels off.

test/fixtures/split-translations/src/split1.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+import React from 'react';
+import {Translate} from 'fusion-plugin-i18n-react';
+
+export default () => <div id="split1-translation"><Translate id="split1"/></div>;
+
```
KevinGrandon: nit: no newline.

build/i18n-manifest-instrumentation-loader.js
KevinGrandon: Nvm, building JS programs in JS. Looks good!

test/fixtures/split-translations/src/root.js
rtsao: Hmm, maybe because this is a fixture

build/compiler.js
```diff
@@ -296,12 +297,8 @@ function getConfig({target, env, dir, watch, cover}) {
                         {pragma},
                       ],
                       cover && require.resolve('babel-plugin-istanbul'),
-                      target === 'node' &&
+                      target === 'web' &&
```
rtsao: This should run in the browser compilation now. Instead of directly instrumenting the server bundle, it passes data to the server compilation out-of-band.

PR #414, opened 2018-07-02
Fix assetUrl when run in the top-level browser scope

KevinGrandon:


entries/client-entry.js
```diff
@@ -34,8 +31,15 @@ Webpack compiles the `__webpack_public_path__ = ...` assignment expression
...
+/* eslint-disable */
 // $FlowFixMe
-__webpack_public_path__ = window.__WEBPACK_PUBLIC_PATH__ + '/'; /* eslint-disable-line */
+__webpack_public_path__ = window.__WEBPACK_PUBLIC_PATH__ + '/';
+
+// $FlowFixMe
+const main = require('__FRAMEWORK_SHARED_ENTRY__');
+/* eslint-enable */
+
+const initialize = main.default || main;
```
KevinGrandon: It's weird, but just copying the logic from the server entry.

KevinGrandon:
!merge

PR #418, opened 2018-07-05
Run user plugins in same order during test as during build

lassedamgaard:
When building user defined babel plugins run first (defined in build/compiler.js) but when testing they run last. This creates inconsistent behaviors and hard to debug issues.


KevinGrandon:
!merge

PR #420, opened 2018-07-06
Do not resolve symlinks during build

kartikssj:
Tools such as `npm link` or `yarn link` use symlinks which when resolved break the build. Disabling resolve.symlinks will ensure they work correctly. (Doc: https://webpack.js.org/configuration/resolve/#resolve-symlinks)

### DIFF
--- a/build/compiler.js
+++ b/build/compiler.js
@@ -356,6 +356,7 @@ function getConfig({target, env, dir, watch, cover}) {
         }),
     ].filter(Boolean),
     resolve: {
+      symlinks: false,
       aliasFields: [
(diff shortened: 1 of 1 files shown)


KevinGrandon (commented):
Thanks for the PR. I'm inclined to give this a shot to see if it improves `yarn link` workflows until we have some bug that proves otherwise. Let me check with the team and get some additional feedback.

rtsao:
Can you be more specific in terms of "break the build"? Perhaps there is another way we can solve the problem. Ideally symlinks should be resolved (hence it is the default behavior with webpack).

PR #428, opened 2018-07-13
Replace Yargs with Sade

swojit:
Fixes https://github.com/uber-web/framework/issues/568 
Improves speed by 15-20x

### DIFF
--- a/bin/cli-runner.js
+++ b/bin/cli-runner.js
@@ -12,32 +12,27 @@ process.on('unhandledRejection', e => {
   throw e;
 });
 
-const fs = require('fs');
-const path = require('path');
-const yargs = require('yargs');
+const sade = require('sade');
 
 module.exports.run = (args /*: any*/) => {
-  const cmdsDir = path.resolve(__dirname, '../commands');
-  const instance = yargs(args)
-    .commandDir(cmdsDir)
-    .demandCommand(1)
-    .version()
-    .help();
-  const cmdPath = path.resolve(cmdsDir, `${instance.argv._}.js`);
-  if (!fs.existsSync(cmdPath)) {
-    const error = new Error(`'${instance.argv._}' is an invalid command\n`);
-    // TODO(#3) ideally, the string from yargs.showHelp() should be part of the error message above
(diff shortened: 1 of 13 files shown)


bin/cli.js
```diff
@@ -10,12 +10,6 @@
...
-  .run(process.argv.slice(2).join(' '))
-  .catch(e => {
-    // eslint-disable-next-line no-console
-    console.error(e.message);
-    if (e.yargs) e.yargs.showHelp();
-    // Rethrow so that there's a non-zero exit code
-    throw e;
-  });
+require('./cli-runner').run(
+  process.argv.map(x => x.split(' ')).reduce((x, y) => x.concat(y))
```
lhorie: Why is this needed?

swojit: It is because of the test files. They pass the arguments like 'dir --test` instead of tokenizing them which breaks sade.

lhorie: I think we should pass Array<string> from the tests to match parameter types under regular usage. As it stands, I'm not sure it does the right thing if we pass something like `--dir="/Program Files/something"`

swojit: It does the right thing. But yes, I was wondering if I should refactor the tests or refactor the cli code for this check. I am pushing another commit with the changes.

bin/cli-runner.js
```diff
@@ -12,32 +12,27 @@ process.on('unhandledRejection', e => {
...
-      value: instance,
-    });
-    return Promise.reject(error);
-  } else {
-    // $FlowFixMe
-    return require(cmdPath).run(instance.argv);
+  const data = require('../commands/index.js');
+  const instance = sade('fusion').version(require('../package.json').version);
+  for (const [command, metadata] of Object.entries(data)) {
+    if (metadata instanceof Object) {
```
angus-c: is this what you want? This will return true for anything that is not a primitive, null or undefined.

swojit:
I needed this for Flow.

bin/cli-runner.js
swojit: I needed this for flow.

PR #429, opened 2018-07-16
Upgrade dependencies to fix release verification

KevinGrandon:


rtsao:
> module/chunk ids in Stats now sort numerical when they are numbers.
Yeah, it looks like the failing tests have hardcoded chunk #s.

PR #433, opened 2018-07-18
Do not log babel notice for assumeNoImportSideEffects

KevinGrandon:
Right now we log a babel warning whenever using `assumeNoImportSideEffects` within a .fusionjs.rc file. This warning can be confusing when not using a custom babel configuration, and only using assumeNoImportSideEffects.


KevinGrandon:
!merge

Issue #426, opened 2018-07-11
Modern bundles for modern browsers


ksheedlo:
### Type of issue

Feature request

### Description

It would be cool if, instead of compiling all ES6 class syntax, async/await etc to ES5, Fusion could detect modern browsers that support these features and send them a more optimized bundle. For instance, class syntax doesn't need to be compiled into functions and prototypes anymore. This could significantly reduce the size of bundles sent to browsers that support modern features, since code volume generally goes up to support the newer syntax in the legacy environment.

Rel #34 

### Current behavior

Fusion delivers code to all browsers using the same compiler settings.

### Desired behavior

Fusion delivers code to modern browsers using more aggressive compiler settings that allow browsers to provide their own class syntax and other ES6 features. Fusion still delivers the same transpiled and polyfilled code to legacy browsers.

gaurav-:
Meteor is probably the first (and so far the only one?) framework to do this properly. Some resources on how this is implemented in Meteor:
- https://github.com/meteor/meteor/blob/devel/History.md#v17-2018-05-28
- https://www.youtube.com/watch?v=vpCotlPieIY&index=4&list=PLTUf4ytkmI8QOopcigiv_dsIubylmfTsX&t=802s
- https://blog.meteor.com/meteor-1-7-and-the-evergreen-dream-a8c1270b0901

PR #434, opened 2018-07-19
Fix i18n discovery on dev restart

rtsao:
We currently have a persistent Babel cache, but translations discovery plugin relies on Babel. Thus in watch mode, when restarting the server, no translations are found because of the cache. This PR sets up a persistent cache that is hydrated on restart to resolve this.

Ideally this isn't necessary, but it is given the current setup with `babel-loader`/`webpack` and Babel caching. Fortunately, this code won't be needed once https://github.com/fusionjs/fusion-cli/issues/426 is implemented because the compilation pipeline will be rather different.

### DIFF
--- a/build/compiler.js
+++ b/build/compiler.js
@@ -473,7 +473,13 @@ function getConfig({target, env, dir, watch, cover}) {
             chunkModuleManifest.set(chunkMap);
           },
         }),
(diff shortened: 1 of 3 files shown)


ganemone:
Can you update the description to describe how this PR fixes it?

rtsao:
!merge

PR #435, opened 2018-07-19
Move globby to deps

lxe:
globby is a dep, not a devDep:

```
build/compiler.js
42:const globby = require('globby');
```


KevinGrandon:
!merge

PR #449, opened 2018-07-25
Add gfx-zopfli

### LINKED ISSUE #111: Replace node-zopfli
The binary deps from this package are annoying to deal with, instead we should use something like
- https://github.com/gfx/universal-zopfli-js
- https://github.com/dfrankland/wasm-zopfli

swojit:
Fixes #111 
After replacing node-nozopfli for its installation issues, Ryan suggested using @gfx/zopfli because it is a fast and universal alternative to the native library.
The behaviour doesn't change at all.
Performance for compression will be slightly slower for gzip compression but given node-zopfli's issues, it's not a dealbreaker.


swojit:
I will definitely keep that in mind. I avoided adding more description because I felt there was nothing more to add after Ryan's description in the referenced issue.
However, for a quick summary, after replacing node-nozopfli for its installation issues, he suggested using @gfx/zopfli because it is a [fast](https://github.com/gfx/universal-zopfli-js#benchmarking) and universal alternative to the native library. 
The behaviour doesn't change at all. 
Performance for compression will be slightly slower for gzip compression but given node-zopfli's issues, it's not a dealbreaker.

mlmorg:
Do we have tests for these compression plugins to make sure that it is actually compressing the files when requested?

mlmorg:
Also, please edit the description with the information you provided in your comment so that we can use it as release notes. Thanks!

swojit:
@mlmorg I am not sure if we have tests for the compression plugins. Let me look that up.

build/compiler.js
```diff
@@ -17,6 +17,7 @@ const ProgressBarPlugin = require('progress-bar-webpack-plugin');
 const webpackDevMiddleware = require('../lib/simple-webpack-dev-middleware');
 const CaseSensitivePathsPlugin = require('case-sensitive-paths-webpack-plugin');
 const {
+  gfxWebpackPlugin,
```
rtsao: I think `zopfliWebpackPlugin` would be a more apt name since it indicates the purpose. It's hard to tell what `gfxWebpackPlugin` based on the name.

lib/compression/index.js
```diff
@@ -9,6 +9,7 @@
 /* eslint-env node */
 
 module.exports = {
+  zopfliWebpackPlugin: require('./gfx-webpack-plugin'),
```
lhorie: `gfx` is the github handle for the wasm binding lib. This file should probably be called zopfli-webpack-plugin

test/cli/build.js
```diff
@@ -433,6 +437,25 @@ test('`fusion build` with assets', async t => {
...
+  fs.readdir(
+    path.resolve(dir, '.fusion/dist/production/client/'),
+    (err, files) => {
+      if (err) throw err;
+      t.ok(files.filter(file => getFileExtension(file) === 'gz'), 'gzip works');
+      t.ok(
+        files.filter(file => getFileExtension(file) === 'br'),
+        'brotli works'
+      );
+      t.ok(files.filter(file => getFileExtension(file) === 'svg'), 'svg works');
```
lhorie: Is this testing SVGO? If so maybe check that the file contents don't have consecutive spaces

test/cli/build.js
```diff
@@ -433,6 +437,25 @@ test('`fusion build` with assets', async t => {
...
 
+test('`fusion build` compresses assets for production', async t => {
+  const dir = path.resolve(__dirname, '../fixtures/compress-assets');
+  await cmd(`build --dir=${dir} --production`);
+
+  fs.readdir(
+    path.resolve(dir, '.fusion/dist/production/client/'),
+    (err, files) => {
+      if (err) throw err;
+      t.ok(files.filter(file => getFileExtension(file) === 'gz'), 'gzip works');
```
lhorie: Maybe just use `path.extname` here

test/cli/build.js
swojit: Or I can just compare file sizes?

lhorie: yeah that works too

swojit: Ah, that makes sense.

PR #451, opened 2018-07-27
Remove redundant custom chunk preloading code

rtsao:
This code was a workaround to allow for async chunks to be preloaded (without being fetched twice) with webpack 3 (Fusion was originally implemented before webpack 4 was released).

The need for this workaround is no longer needed as it has been fixed in webpack 4 ([this is the relevant PR](https://github.com/webpack/webpack/pull/3333)).


rtsao:
Yep, the information gathering part is still all there, this plugin was just for making the webpack 3 runtime able to see if some async chunks were already preloaded (which isn't needed anymore since this works out of the box in webpack 4)

KevinGrandon:
LGTM then, happy to review once tests are passing.

rtsao:
!merge

PR #454, opened 2018-07-30
Remove forked file-loader dependency

rtsao:
These are tightly coupled and this will make later refactoring easier


rtsao:
!merge

PR #462, opened 2018-08-02
Display application running URL on yarn dev

samundrak:
Will log info of URL where the application is running only if `--open=false` command is provided,
mostly it helps us to not guess the PORT where the server is running.

### DIFF
--- a/commands/dev.js
+++ b/commands/dev.js
@@ -60,6 +60,9 @@ exports.run = async function(
         // $FlowFixMe
         testRuntime ? testRuntime.run() : Promise.resolve(),
       ]);
+      if (!open) {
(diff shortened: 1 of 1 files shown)


commands/dev.js
```diff
@@ -60,6 +60,9 @@ exports.run = async function(
         // $FlowFixMe
         testRuntime ? testRuntime.run() : Promise.resolve(),
       ]);
+      if (!open) {
```
KevinGrandon: Thanks for the feature addition. I wonder if we should just show this message all the time, whether or not it opened.

samundrak: I also think showing message all time will be better in case accidentally one closes tab then they can always look on the terminal, please let me know if i should update PR

PR #473, opened 2018-08-15
Make react/enzyme dependency soft

lhorie:
Currently fusion-cli depends on react and enzyme in order to be used at all. This was originally done to make it easier to test React, but the overall goal of fusion-cli is to be a CLI+compilation infrastructure that is agnostic of view libraries.

This PR allows fusion-cli to be installed and used without throwing react/enzyme-related errors, while still enabling them to be used if they're detected.


build/server-error.js
```diff
@@ -8,15 +8,7 @@
 
 /* eslint-env node */
 
-const React = require('react');
-const RedBox = require('redbox-react').RedBoxError;
```
KevinGrandon: If we're getting rid of this, we should remove the `redbox-react` dependency from package.json

lhorie: Yeah, good catch!

PR #477, opened 2018-08-20
Prevent tree-shaking of polyfills

rtsao:
If `assumeNoImportSideEffects` is set to `true`, then imports for side effects will be tree-shaken (per https://github.com/webpack/webpack/issues/6571 effectful imports are not considered side effect free by webpack).

Because the API of core-js relies on import side effects (rather than some actual polyfill method) this means polyfills are erroneously never included.

This PR avoids this behavior by changing the polyfill imports to use `require` instead.


rtsao:
!merge

PR #482, opened 2018-08-24
Revert require.cache busting

lhorie:
- [clearing require.cache](https://github.com/fusionjs/fusion-cli/blob/fa1ed5fac5624eefdb8d5b9103ee4436885cc386/build/jest/jest-transformer.js#L15) was a workaround to prevent test failures when running server and browser tests concurrently. However, this workaround causes sporadic `Module did not self-register` errors from binary dependencies, which themselves cannot be worked around.
- original issue was isolated in https://github.com/facebook/jest/issues/6887
- new workaround: use `fusion test --env node && fusion test --env jsdom` instead of `fusion test` if you have server and browser tests running concurrently
- downsides: coverage for node and jsdom tests are not consolidated

Next steps:

- tell people to use the new workaround
- wait for upstream fix in Jest

### DIFF
--- a/build/jest/base-jest-config.js
+++ b/build/jest/base-jest-config.js
@@ -33,7 +33,6 @@ function getReactSetup() {
 const reactSetup = getReactSetup();
 
 module.exports = {
-  cache: false,
   coverageDirectory: `${process.cwd()}/coverage`,
   coverageReporters: ['json'],
   rootDir: process.cwd(),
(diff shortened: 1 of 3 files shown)


mkrishnaatuber:
Is there any workaround for this?

KevinGrandon:
> Is any permanent fix in progress for this? Otherwise our project test coverage is constantly reported wrongly.

Yes, this was fixed in a recent version of fusion-cli, and should exist in 1.10.2.

Fix was here: https://github.com/fusionjs/fusion-cli/commit/005516ea412a462e5dea46c70797cc6fa4c2c6c7

PR #489, opened 2018-08-29
Upgrade dependencies

rtsao:
Supercedes https://github.com/fusionjs/fusion-cli/pull/481

It appears upgrading babel dependencies from `v7.0.0-rc.1` to `v7.0.0` made it so that the coverage objects in `coverage-final.json` as outputted by Jest are sometimes (but not always) nested inside an object in its "data" property. This fixes the failing test which was also in #481


build/convert-coverage.js
```diff
@@ -48,9 +48,10 @@ module.exports = function(rootDir /*: string */) {
     [rootDir].forEach(dir => {
       // $FlowFixMe
       const coverage = require(`${dir}/coverage/coverage-final.json`);
-      Object.keys(coverage).forEach(filename =>
-        map.addFileCoverage(coverage[filename])
-      );
+      Object.keys(coverage).forEach(filename => {
```
KevinGrandon: Can we add some comment explaining why this is terrible?

PR #490, opened 2018-08-30
Add name for each environment to jest config

KevinGrandon:
Full write-up on why this is required here: https://github.com/facebook/jest/issues/6887#issuecomment-417170450


KevinGrandon:
!merge

PR #501, opened 2018-09-04
Add no-cache flag when --debug is set

KevinGrandon:
This is required for VSCode users to be able to debug from within VSCode.

See: https://github.com/facebook/jest/issues/6683#issuecomment-417515963


KevinGrandon:
!merge

PR #502, opened 2018-09-04
Support a whitelist of Jest CLI arguments

### LINKED ISSUE #500: Support all Jest CLI onlyChanged option
### Type of issue

Feature request

### Description

Should be able to run tests with any of the jest CLI options. It appears with `fusion test --watch`, the `--watch` flag is passed through, but others aren't.

I'm specifically looking for `fusion test --onlyChanged` to work.

### Environment

fusion-cli@1.8.3

KevinGrandon:
Adds a whitelist of supported Jest CLI arguments which seem reasonable and useful. This does not add arguments which have potential to change the convention of test naming, coverage collection, or environment.

Sade output looks like:
image

Fixes #500

### DIFF
--- a/commands/test.js
+++ b/commands/test.js
@@ -8,6 +8,7 @@
 
 /* eslint-env node */
 
+const {allowedJestOptions} = require('../build/jest/cli-options');
 const {TestAppRuntime} = require('../build/test-runtime');
 
 exports.run = async function(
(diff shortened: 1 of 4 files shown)


commands/test.js
```diff
@@ -13,30 +13,66 @@ const {TestAppRuntime} = require('../build/test-runtime');
...
-    updateSnapshot,
-    coverage,
     configPath,
     // Allow snapshots to be updated using `-u` as well as --updateSnapshot.
     // We don't document this argument, but since jest output automatically
     // suggests this as a valid argument, we support it in case it's used.
     u,
+    updateSnapshot,
+    // Arguments which are passed through into jest
+    ...rest
```
lhorie: IIRC, we need to also explicitly specify arguments in index.js since we now use sade

KevinGrandon: Seems like it's only for documentation, or when we run `--help`, but we can list them there as well!

KevinGrandon:
!merge

PR #486, opened 2018-08-27
Refactor compiler

rtsao:
### Summary
- Replaces vanilla babel-loader with simplified version which also caches static analysis metadata in lockstep with the transformed code. This eliminates the current brittleness with translations discovery (related to persistent cache invalidation) and removes the need for hacky workaround fixes in https://github.com/fusionjs/fusion-cli/pull/442 and https://github.com/fusionjs/fusion-cli/pull/434.
- Replaces our forked file-loader with a vastly simpler version that doesn't rely on singletons or shared state.
- Replaces chunk id loader with a version that doesn't rely on singletons.
- Simplifies Babel configuration
- Deletes a considerable amount of vestigial code (mostly relating to the original evergreen bundles implementation and replaced test functionality)
- Adds proper dev mode configuration to react babel preset.

**Webpack compilation/loader communication**

The old pattern relied on shared state via singletons which was far from ideal. This has been replaced with usage of loader context (and plugins that provide/consume this context). Shared state  (between compilations and loaders as well as between sibling compilations) is now instantiated, passed, and mutated in a much more explicit manner.

<hr/>

Resolves #7
Resolves #8 
Resolves #9
Resolves #15 
Resolves #47 

Lays the groundwork for #393 #426 https://github.com/fusionjs/rfcs/pull/7 https://github.com/fusionjs/rfcs/pull/10

PR #505, opened 2018-09-05
Add @babel/plugin-transform-flow-strip-types in favor of @babel/preset-flow

KevinGrandon:
* Installs and uses `@babel/plugin-transform-flow-strip-types`.
* Introduces a test case for a failure after upgrading to fusion-cli@1.10.0.

Note: We were including `@babel/preset-flow` inside of our `get-babel-config.js` file, but it seems to not actually work. It seems that this is due to some other plugins running before the preset, possibly `@babel/plugin-proposal-class-properties`. We have to unshift the plugin to strip flow types first. This means that we can remove the `@babel/preset-flow` plugin.

This bug was introduced from https://github.com/fusionjs/fusion-cli/pull/486

### DIFF
--- a/build/babel-plugins/babel-plugin-utils/test/index.js
+++ b/build/babel-plugins/babel-plugin-utils/test/index.js
@@ -40,7 +40,9 @@ test('with flow types', t => {
     import type {footype} from 'bar';
     let baz: string = foo();
   `,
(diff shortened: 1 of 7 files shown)


KevinGrandon:
!merge

PR #516, opened 2018-09-14
Fix Flow-annotated class properties/methods

rtsao:
Babel 7.0.0 shipped a regression:

See: https://github.com/babel/babel/issues/8417, https://github.com/babel/babel/pull/8584

This PR uses a fork that applies the fix https://unpkg.com/@rtsao/plugin-proposal-class-properties@7.0.1-patch.1/lib/index.js


rtsao:
!merge

PR #518, opened 2018-09-14
Fix missing translations in dev after rebuild

rtsao:
All translations were being erroneously evicted on invalidation. Instead, only translations from modified files should be evicted (which will be re-analyzed later).

Our tests failed to cover this case because only translations from the file being modified were checked. This PR adds a regression test to check the rendered translation within an unmodified file after a rebuild.


rtsao:
!merge

PR #524, opened 2018-09-19
Fix polyfilling with sideEffectFree option and add regression test

rtsao:
Side effect free require pruning was added to webpack so the existing CommonJS workaround stopped working.

This PR adds a regression test to prevent future breakages


rtsao:
!merge

PR #527, opened 2018-09-21
Add experimentalCompile flag

ganemone:
The `experimentalCompile` flag will tell fusion to run babel with all plugins against every bundled file. After we confirm this approach is viable, we will remove this flag and make it the default behavior in a breaking change. Among other things, this allows projects to have source files outside of `src` with no additional configuration which is useful when using fusion in a monorepo.


ganemone:
!merge

PR #528, opened 2018-09-21
Use require.resolve rather than requiring modules directly

ganemone:
Webpack serializes these anyway, so it is simpler to pass the paths directly.


ganemone:
!merge

Issue #393, opened 2018-06-14
Revisit chunk hashing / long term caching


rtsao:
See: https://github.com/fusionjs/fusion-cli/pull/385

PR #531, opened 2018-09-25
Fix chunk hashing, refactor JS asset code

rtsao:
- Move `__webpack_public_path__` code into their own files. This is less error prone because it is guaranteed to come first before other code (because those files are placed first in the entries array), so no need to worry about hoisted imports, etc.
- Configure webpack to use a runtime chunk which is required for long-term caching of chunks. This runtime chunk is invalidated for every change but is tiny; the remaining chunks only change when modified.
- Add md5 hashing of chunks, which makes hashes deterministic based on file contents. Without this, hashes would not change if uglify settings were changed (for example) See: https://github.com/webpack/webpack/issues/4659 for more info
- Enforce that the server bundle has only one chunk
- Register RoutePrefixToken, which will be consumed by newer packages instead of the context property. This token should be registered in fusion-cli because the prefix stripping code also lives in fusion-cli. Eventually we should move probably move all route prefixing code into a separate plugin.
- Register CriticalChunkIds, which will be consumed by newer versions of fusion-react, etc.
- Register SSRBodyTemplate, which will replace the legacy template used in fusion-core. For backwards compat, this template uses both the legacy `ctx.preloadChunks` as well as the new `CriticalChunkIds` token.
- Remove preload hints, which won't be compatible with serving both `<script type="module">` and `<script>` bundles.

Fixes https://github.com/fusionjs/fusion-cli/issues/393

### DIFF
--- a/plugins/ssr-plugin.js
+++ b/plugins/ssr-plugin.js
@@ -6,8 +6,11 @@
  * @flow
  */
 /* eslint-env node */
+/* global __webpack_public_path__ */
 
 /*::
+declare var __webpack_public_path__: string;
+
 import type {
   Context,
   SSRBodyTemplate as SSRBodyTemplateService,
@@ -19,13 +22,15 @@ import {
   escape,
   consumeSanitizedHTML,
   CriticalChunkIdsToken,
+  RoutePrefixToken,
 } from 'fusion-core';
 
 const SSRBodyTemplate = createPlugin({
   deps: {
     criticalChunkIds: CriticalChunkIdsToken.optional,
+    routePrefix: RoutePrefixToken.optional,
   },
-  provides: ({criticalChunkIds}) => {
+  provides: ({criticalChunkIds, routePrefix}) => {
     return ctx => {
       const {htmlAttrs, bodyAttrs, title, head, body} = ctx.template;
       const safeAttrs = Object.keys(htmlAttrs)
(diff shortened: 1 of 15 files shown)


test/cli/build.js
```diff
@@ -167,11 +167,11 @@ test('`fusion build` app with dynamic imports chunk hashing', async t => {
     'vendor file hash should not change'
   );
   // TODO(#393) Add this back https://github.com/fusionjs/fusion-cli/pull/385
```
KevinGrandon: Nit: remove comment.

test/cli/build.js
```diff
@@ -167,11 +167,11 @@ test('`fusion build` app with dynamic imports chunk hashing', async t => {
...
-  // t.equal(
-  //   distFiles.clientMainFile,
-  //   rebuiltDistFiles.clientMainFile,
-  //   'main file hash should not change'
-  // );
+  t.equal(
```
KevinGrandon: 💯

plugins/ssr-plugin.js
```diff
@@ -0,0 +1,112 @@
...
+      // $FlowFixMe
+      const safeHead = head.map(consumeSanitizedHTML).join('');
+      // $FlowFixMe
+      const safeBody = body.map(consumeSanitizedHTML).join('');
+
+      const coreGlobals = [
+        `<script nonce="${ctx.nonce}">`,
+        `window.performance && window.performance.mark && window.performance.mark('firstRenderStart');`,
+        routePrefix && `__ROUTE_PREFIX__ = ${JSON.stringify(routePrefix)};`,
+        `__FUSION_ASSET_PATH__ = ${JSON.stringify(__webpack_public_path__)};`, // consumed by fusion-clientries/client-entry
```
KevinGrandon: nit: typo in path.

test/cli/build.js
rtsao: For some reason I was thinking the bot would make me keep this comment, but that isn't the case. Will remove.

rtsao:
!merge

PR #536, opened 2018-10-02
Ensure runtime chunk is last in legacy ctx.syncChunks

rtsao:


rtsao:
Actually, I think this is unnecessary

PR #541, opened 2018-10-03
Remove vestigial webpack-dev-server config

rtsao:
We don't use webpack-dev-server anymore so this configuration is not used by anything.


rtsao:
!merge

PR #543, opened 2018-10-03
Update snapshot command in test

KevinGrandon:
Seems this is necessary for release verification. I'm not sure why, but we've had problems with node -e in the past.


KevinGrandon:
Going to just land this to get release verification unblocked. Feel free to audit.

PR #545, opened 2018-10-03
Fix incorrect sourcemap filenames

rtsao:
webpack-plugin-hash-output was not also renaming source map files. This meant that the source map files had a different basename than the JS chunks.

Removing for now until this is fixed.

Note: this means changing uglify settings will yield hash collisions. So we should put back this behavior once the problem is addressed.


rtsao:
!merge

PR #548, opened 2018-10-04
Fix client-side RoutePrefixToken registration

rtsao:
This isn't used anywhere yet, but the actual value needs to be registered (just like on the server)


rtsao:
!merge

PR #550, opened 2018-10-04
Fix client-side token registration error

rtsao:
This PR resolves a client-side error from a token being registered without being used and adds a regression test


rtsao:
!merge

PR #552, opened 2018-10-04
Fix assetUrl in dev in certain instances

lhorie:
Rationale:

In some cases, webpack was attempting to compile the path `../client/[file]` as `/client/[file]` (due to path normalization running [this logic](https://github.com/webpack/memory-fs/blob/master/lib/normalize.js#L33) instead of [this logic](https://github.com/webpack/memory-fs/blob/master/lib/normalize.js#L54))

This PR makes that case work in dev, while avoiding regressions in production

### DIFF
--- a/build/compiler.js
+++ b/build/compiler.js
@@ -40,7 +40,7 @@ const {DeferredState} = require('./shared-state-containers.js');
 const {
   translationsManifestContextKey,
   clientChunkMetadataContextKey,
-  modeKey,
+  devContextKey,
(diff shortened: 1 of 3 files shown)


lhorie:
I tried but for the life of me I couldn't reproduce the issue outside of the repo where it originally manifested itself

rtsao:
Hmm, I wonder if this has to do with lockfile differences (likely webpack?)

rtsao:
!merge

PR #555, opened 2018-10-05
WIP - Add test case for proxy SSR error caching

KevinGrandon:


test/fixtures/server-startup-control/delayed-compilation.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+const {execSync} = require('child_process');
+
+module.exports = () => {
+  return {
+    visitor: {
+      StringLiteral(path /* : any */) {
+        if (path.node.value === 'TRIGGER-BABEL-DELAY') {
+          console.log('Triggered babel delay');
+          execSync('sleep 3');
```
KevinGrandon: Should probably use node-IPC or something for more robust/faster tests, but this is a lot less code. We can update this if we see flakiness on CI.

build/dev-runtime.js
```diff
@@ -27,6 +27,7 @@ function Lifecycle() {
   return {
     start: () => {
       state.started = true;
+      state.error = undefined;
```
KevinGrandon: 💥

KevinGrandon: All that testing just to find this 😂😂😂

test/cli/dev.js
```diff
@@ -481,3 +481,77 @@ test('`fusion dev` app with split translations integration (cached)', async t =>
 
   t.end();
 });
+
+test('`fusion dev` proxy gracefully recovers from cached SSR errors', async t => {
+  const dir = path.resolve(__dirname, '../fixtures/server-startup-control');
+
+  async function waitForCompileToStart() {
+    // Should figure out a better heuristic for this.
```
rtsao: Would this strategy work for your use case?

https://github.com/fusionjs/fusion-cli/blob/bd180b7ec79ed45cd475f4159d28421d668a2b01/test/cli/dev.js#L404-L417

build/dev-runtime.js
ganemone: Nice find.

PR #557, opened 2018-10-08
Fix skipped transpilation of async functions in certain instances

rtsao:
https://github.com/fusionjs/fusion-cli/commit/61f61d42fff0fd44a557b009b1ddd32d750cf22b surfaced an issue where async functions were not being transpiled in certain instances. One example is async function usage where:
- Must be default export (non-default export works fine)
- Must be in object in `createPlugin` call (seems to work for other functions)

This PR adds a regression test and fixes transpilation of async functions in these instances.

I can't seem to get fast-async to work correctly anymore, so this PR removes it (replacing it with the normal regenerator transform).


build/get-babel-config.js
```diff
@@ -95,10 +95,7 @@ module.exports = function getBabelConfig(opts /*: BabelConfigOpts */) {
     envPresetOpts.targets = {
       ie: 9,
     };
-    envPresetOpts.exclude = [
-      'transform-regenerator',
```
rtsao: It seems both exclusions was causing trouble.

rtsao:
!merge

PR #559, opened 2018-10-09
Add gql macro from fusion-apollo

ganemone:
Adds support for loading graphql files via gql macro from fusion-apollo.

Example:

```js
import {gql} from 'fusion-apollo';

const query = gql('./query.gql');
const schema = gql('./schema.gql');
```

### DIFF
--- a/build/babel-plugins/babel-plugin-gql/test/fixtures/expected-import-destructuring
+++ b/build/babel-plugins/babel-plugin-gql/test/fixtures/expected-import-destructuring
@@ -1,0 +1,9 @@
+import { gql } from 'fusion-apollo';
+import { gql as gqlOther } from 'gql';
+const path = './test';
+
+require("__SECRET_GQL_LOADER__!./path");
+
+gqlOther(path);
+
+require("__SECRET_GQL_LOADER__!./path");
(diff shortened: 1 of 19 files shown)


build/babel-plugins/babel-plugin-gql/test/fixtures/expected-import-destructuring
```diff
@@ -0,0 +1,9 @@
+import { gql } from 'fusion-apollo';
```
rtsao: I think fusion-apollo needs a named gql export, otherwise this may cause errors.

Ideally this is something like:

```js
export const gql = (() => {
  throw new Error("gql called at runtime");
} /*: string => string */);
```

Then $FlowFixMe can be removed too.

ganemone: Yea, I was going to do that next.

ganemone: The $FlowFixMe is actually there because we are not installing fusion-apollo

ganemone:
!merge

PR #561, opened 2018-10-09
Fix test gitignores

rtsao:
https://github.com/fusionjs/fusion-cli/commit/452369b9c337665a206e2ac48b38d3499910d893 introduced a gitignore change accidentally un-ignored some babel cache files.

This PR reverts the root gitignore change and adds local gitignore to preserve the fixture node_modules.


rtsao:
!merge

Issue #562, opened 2018-10-10
fusion dev --dir <directory> hiccups


scott-hornberger:
### Description
When running `fusion dev --dir <directory>`, I get the error:
```
Error: No entry found at /path/to/project/<directory>/<directory>/.fusion/dist/development/server/server-main.js
    at [eval]:40:20
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at Object.runInThisContext (vm.js:139:38)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:652:30)
    at evalScript (bootstrap_node.js:466:27)
    at startup (bootstrap_node.js:167:9)
    at bootstrap_node.js:612:3
```
It seems to build fine, but it can't find `server-main.js` because it's looking for `<directory>/<directory>` instead of `<directory>`. 

### Current behavior
It fails to find `/path/to/project/<directory>/<directory>/.fusion/dist/development/server/server-main.js` and will fail to find other resources because of the hiccup.

### Expected behavior
It should use  `/path/to/project/<directory>/.fusion/dist/development/server/server-main.js` and find all resources at the directory.

### Steps to reproduce

1. Scaffold a new project.
2. Move everything besides `package.json` and `node_modules` to a sub folder.  
3. Run `fusion dev --dir <subfolder>

### Your environment

* fusion-cli version: fusion, 1.11.5

* Node.js version: v8.11.4

* npm version: 5.6.0

* Operating System: Mac OS X

scott-hornberger:
I forgot to mention: If I put a recursive symlink `<directory>` in `<directory>` that points to `<directory>` then it runs successfully!

PR #565, opened 2018-10-12
Fix `fusion dev --dir`

lhorie:
Fixes https://github.com/fusionjs/fusion-cli/issues/562

### DIFF
--- a/test/cli/dev.js
+++ b/test/cli/dev.js
@@ -32,11 +32,12 @@ test('`fusion dev` works', async t => {
   t.end();
 });
 
-test.only('`fusion dev --dir` works w/ relative dir', async t => {
+test('`fusion dev --dir` works w/ relative dir', async t => {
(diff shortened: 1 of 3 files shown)


test/cli/dev.js
```diff
@@ -32,6 +32,25 @@ test('`fusion dev` works', async t => {
   t.end();
 });
 
+test('`fusion dev --dir` works w/ relative dir', async t => {
```
rtsao: I think relative dir should be explicitly disallowed. Or alternatively, it should be immediately converted to an absolute path as soon as it is provided

rtsao: In our own code, we should be able to assume dir is an absolute path and as a replacement for cwd.

lhorie: From what I could tell, we do convert it to a absolute path when we spawn the child process in the dev runtime, but it's common for people to pass in a relative path to the CLI arg

PR #566, opened 2018-10-12
Fix SSR plugin render CSP nonce

ksheedlo:
This change fixes a bug where CSP nonces were not included in the generated script tags, which could cause CSP to block the scripts.


KevinGrandon:
!merge

PR #570, opened 2018-10-16
Fix issue with babel blacklist regexes

ganemone:
The babel blacklist regexes were matching anything prefixed with
react. This meant we were unknowingly blacklisting all react components
from getting babelified. This updates the regexes and adds a regression test.

### DIFF
--- a/build/compiler.js
+++ b/build/compiler.js
@@ -246,11 +246,11 @@ function getConfig({target, env, dir, watch, state}) {
           test: /\.jsx?$/,
           exclude: [
             // Blacklist mapbox-gl package because of issues with babel-loader and its AMD bundle
-            /node_modules\/mapbox-gl(\/.*?)*$/,
+            /node_modules\/mapbox-gl\//,
             // Blacklist known ES5 packages for build performance
-            /node_modules\/react-dom(\/.*?)*$/,
-            /node_modules\/react(\/.*?)*$/,
-            /node_modules\/core-js(\/.*?)*$/,
+            /node_modules\/react-dom\//,
+            /node_modules\/react\//,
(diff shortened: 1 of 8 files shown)


build/compiler.js
```diff
@@ -246,11 +246,11 @@ function getConfig({target, env, dir, watch, state}) {
...
           exclude: [
             // Blacklist mapbox-gl package because of issues with babel-loader and its AMD bundle
-            /node_modules\/mapbox-gl/,
+            /node_modules\/mapbox-gl(\/.*?)*$/,
             // Blacklist known ES5 packages for build performance
-            /node_modules\/react-dom/,
-            /node_modules\/react/,
-            /node_modules\/core-js/,
+            /node_modules\/react-dom(\/.*?)*$/,
+            /node_modules\/react(\/.*?)*$/,
```
rtsao: I think these are matched against actual file paths (not directories), so this should suffice, right?

```suggestion
            /node_modules\/react\//,
```

test/fixtures/custom-babel/plugin.js
```diff
@@ -1,9 +1,10 @@
 module.exports = () => ({
   visitor: {
     StringLiteral(path) {
-      if (path.node.value === "helloworld") {
-        path.node.value = "transformed_helloworld_custom_babel";
+      if (path.node.value === 'helloworld') {
+        console.log('TRIGGERING...');
```
rtsao:
```suggestion

```

build/compiler.js
ganemone: Ok - Will update

ganemone:
!merge

PR #572, opened 2018-10-18
Add specialized build for modern browsers

rtsao:
Add compilation of a modern JS build for modern browsers. By default, this is skipped in the `dev` command, but can be turned on with `--forceLegacyBuild`.


rtsao:
!merge

PR #573, opened 2018-10-18
Update gql loader to use gql tag

ganemone:
This will allow us to treat all graphql code consistently by
always using a parsed ast.

### DIFF
--- a/build/loaders/gql-loader.js
+++ b/build/loaders/gql-loader.js
@@ -10,7 +10,10 @@
 module.exports = function gqlLoader(content /*: string */) {
   // NOTE: For now, we are simply loading queries and schemas as strings.
   // However, we may wish to load a pre-parsed graphql AST, similar to how https://github.com/samsarahq/graphql-loader works.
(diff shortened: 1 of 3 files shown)


rtsao:
I think this needs to be rebased on top of https://github.com/fusionjs/fusion-cli/commits/master

build/loaders/gql-loader.js
```diff
@@ -10,7 +10,10 @@
 module.exports = function gqlLoader(content /*: string */) {
   // NOTE: For now, we are simply loading queries and schemas as strings.
   // However, we may wish to load a pre-parsed graphql AST, similar to how https://github.com/samsarahq/graphql-loader works.
-  const result = `module.exports = ${JSON.stringify(content.toString())};`;
+  const result = `
+  const gql = require('graphql-tag');
+  module.exports = gql(${JSON.stringify(content.toString())});
```
rtsao: Is the result of graphql-tag serializable? Could this be run at build-time?

ganemone: It might be - long term I do want to investigate that.

PR #576, opened 2018-10-18
Ensure nonces are set for all scripts (including dynamically loaded scripts)

rtsao:
Previously, dynamically loaded scripts did not have the appropriate nonce attribute, which could lead to CSP errors.

Also, this ensures the new script loader in https://github.com/fusionjs/fusion-cli/pull/572 sets these as well.


ksheedlo:
👍

rtsao:
!merge

PR #579, opened 2018-10-18
Set script nonce attribute rather than property

rtsao:
Fixes an oversight in https://github.com/fusionjs/fusion-cli/pull/576 which was mistakenly setting the element property instead of the attribute.


rtsao:
!merge

PR #582, opened 2018-10-19
Load modern bundles as regular scripts

rtsao:
There seems to be crossorigin credentials caveats with module scripts. Adding crossorigin attributes will break Safari (known bug), but is needed in other browsers also.


rtsao:
!merge

PR #584, opened 2018-10-22
Release v1.13.0-beta.4

ganemone:


KevinGrandon:
I'm going to just press the merge button here. It seems that webhook deliveries were failing on Github, and we'd need to re-trigger to get passing CI. Should be fine to just land this pre-release though.

https://status.github.com/messages

PR #588, opened 2018-10-23
Fix issue with coverage and gql macro

ganemone:
This fixes an issue with the inline gql macro and jest coverage
instrumentation. Also updates the plugins to read the graphql files
during test at runtime to resolve potential caching issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules/
 .fusion/
 .nyc_output/
 yarn-error.log
+coverage/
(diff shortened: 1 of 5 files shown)


rtsao:
Is there a need for it to be an absolute path?

```js
path.relative(sourceFilePath, gqlFilePath);
```

ganemone:
> Is there a need for it to be an absolute path?
> 
> ```js
> path.relative(sourceFilePath, gqlFilePath);
> ```

I think it would need to be a relative path from `cwd`, not from the file. But I could use `__dirname`

ganemone:
!merge

PR #593, opened 2018-10-26
Force class transpilation in modern bundle to workaround bugs in Edge

rtsao:
Classes need to be transpiled in the modern bundle to work around a couple bugs with Edge with regard to `super()`


rtsao:
We could potentially have a runtime check to determine if the browser is Edge or not, but this makes it a bit more complex and potentially flaky.

rtsao:
!merge

ksheedlo:
how much is this going to cost in terms of the size of our modern bundles?

Issue #533, opened 2018-09-27
Open up direct modification of Jest's testMatch option.


alee8:
### Type of issue

fusion.js, through fusion-cli, does not allow direct specification of jest's `testMatch` configuration option.

### Description

Currently, fusion.js allows indirect customization of `testMatch` (i.e., `**/${testFolder}/**/*.js`) by modifying the `testFolder` option through the fusion-cli `testFolder` option.  This configuration does not support tests outside of `testFolder` (e.g., alongside the files being tested). It is not possible to differentiate tests files from files being tested by adding `test|spec` to the test file name. Finally, Jest's `testMatch` supports an array of glob patterns that are not possible in fusion.js' tests.

### Current behavior

- Test files are located in `testFolder`
- Test filenames are not differentiated by file names that conform to a naming pattern (e.g., file-test.js, file.test.js, file-spec.js, file-spec.js)
- JS fixture/mock files in `testFolder` cannot be excluded as part of the `testMatch` glob pattern.

### Expected behavior

- Support tests in files that are located in `testFolder`.
- Support test filenames that follow a specific naming pattern to hone in on enumeration of test files.
- Support test files that are colocated with files being tested and not in a `testFolder`

### Your environment

* fusion-cli version: 1.9.0
* Node.js version (`node --version`): 8.11.4
* npm version (`npm --version`): 5.6.0
* Operating System: MacOS High Sierra

PR #596, opened 2018-10-29
Add testMatch/testRegex options to `fusion test`

lhorie:
Fixes https://github.com/fusionjs/fusion-cli/issues/533

Some projects at Uber have large existing test suites structured differently than the-jest-way(tm). This PR allows those users to configure their project to find tests where they currently exist in order to avoid expensive refactors.

This also helps pave the road for monorepo configuration

### DIFF
--- a/build/jest/base-jest-config.js
+++ b/build/jest/base-jest-config.js
@@ -8,7 +8,10 @@
 
 /* eslint-env node */
 
-const testFolder = process.env.TEST_FOLDER || '__tests__';
+const matchField = process.env.TEST_REGEX ? 'testRegex' : 'testMatch';
+const matchValue = process.env.TEST_FOLDER
+  ? [`**/${process.env.TEST_FOLDER || '__tests__'}/**/*.js`]
+  : process.env.TEST_REGEX || JSON.parse(process.env.TEST_MATCH);
 
 function getReactVersion() {
   // $FlowFixMe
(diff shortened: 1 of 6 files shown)


build/jest/base-jest-config.js
```diff
@@ -8,7 +8,10 @@
 
 /* eslint-env node */
 
-const testFolder = process.env.TEST_FOLDER || '__tests__';
+const matchField = process.env.TEST_REGEX ? 'testRegex' : 'testMatch';
+const matchValue = process.env.TEST_FOLDER
+  ? [`**/${process.env.TEST_FOLDER || '__tests__'}/**/*.js`]
```
KevinGrandon: So we're removing the `__tests__` default folder, correct? Will this start running tests on files we had previously expected it not to? `__integration__` for example? My worry is that it might, and that we probably don't have any test cases to catch this.

lhorie: Default now uses TEST_MATCH with the same value as before. We do have tests w/ default config, as well as overridden testFolder (for integration), and now for overridden testMatch and testRegex.

The final jest config object under default fusion-cli configuration is still `testMatch: [**/__tests__/**/*.js]` as it was before

KevinGrandon: Gotcha, didn't see that default setting in the CLI spec.

Issue #547, opened 2018-10-04
Differences in asset serving in prod vs dev


rtsao:
https://github.com/fusionjs/fusion-cli/pull/130 is interesting. This was added to allow user plugins to set headers on assets. 

But it's worth noting in development we serve assets from a separate process (in the dev proxy, serving assets from memory instead) so in dev, the app will never receive these requests, but in prod they will.

We should probably resolve this inconsistency. It seems there are a few potential options:
- Make asset serving always happen before reaching the user app request lifecycle.
- In development, write assets to disk in development and use the normal asset middleware in the request lifecycle. 
- Remove asset serving from the dev proxy, instead let the requests go to the normal asset middleware, but instead of reading from disk, proxy to some other server that serves the in-memory assets

PR #597, opened 2018-10-29
Fix server-side only assets in development

rtsao:
This PR changes assets serving in development to match how it works in production. Assets are no longer served from memory.

Fixes an issue with server-side assets in development not being served correctly. Due to a mistake in the test fixture for this (fixed in this PR), tests were erroneously passing.

Also resolves https://github.com/fusionjs/fusion-cli/issues/547


test/fixtures/assets/src/main.js
```diff
@@ -3,13 +3,13 @@ import App from 'fusion-core';
...
 import * as jsonData from './static/test.json';
 
+import serverAsset from './server-asset.js';
+
 const hoistedUrl = assetUrl('./static/test.css');
 if (typeof window !== 'undefined') {
   window.__hoistedUrl__ = hoistedUrl;
 }
 
-import {serverAsset} from "./server-asset.js";
```
KevinGrandon: Whoa, this file just didn't exist before? 😮

rtsao: It existed, but was empty 😞

rtsao:
!merge

PR #604, opened 2018-11-01
Improve DX of testMatch

lhorie:
- throw if multiple redundant configs are set (testMatch/testRegex/testFolder)
- change testMatch to take comma-separated list, vs JSON

### DIFF
--- a/commands/test.js
+++ b/commands/test.js
@@ -39,6 +39,16 @@ exports.run = async function(
     }
   });
 
+  if (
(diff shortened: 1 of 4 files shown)


commands/test.js
```diff
@@ -39,6 +39,16 @@ exports.run = async function(
     }
   });
 
+  if (
```
angus-c: `[testFolder, testMatch , testRegex ].filter(s => s !== '').length > 1` ?

PR #605, opened 2018-11-01
Fix race condition in script loading

rtsao:
Because dynamically inserted scripts are implicitly async by default (which takes precedence over defer), a race condition was introduced where bundle scripts might be executed early and/or out-of-order.

This change forces these scripts to be explicitly not async, so the defer attribute will work.


rtsao:
!merge

PR #610, opened 2018-11-07
Add bundling of service worker entry

rtsao:
[RFC](https://github.com/fusionjs/rfcs/blob/sw/text/0000-service-worker.md)


rtsao:
cc @angus-c

angus-c:
Seems good but I'm still uncertain where to get the assets from w/r/t to `ctx`.

In [WIP service worker plugin PR](https://github.com/fusionjs/fusion-plugin-service-worker/blob/for-new-template/src/server.js#L20) — I'm currently using `ctx.chunkUrlMap` for both precachePaths and cacheablePaths. 

I'm assuming precachePaths (i.e. core bundles—vendor, main, runtime) and cacheablePaths (all JS bundles) can be derived at build time (with base asset path modified at runtime)

rtsao:
Yeah, JS asset names could made into a virtual module (and imported in the sw entry). Alternatively, this information could be registered under a well-known token by fusion-cli and depended on via the DI system. Or possibly something else depending on how https://github.com/fusionjs/rfcs/pull/7 shakes out. But whatever this ends up being shouldn't have any impact on anything this PR is concerned with.

PR #611, opened 2018-11-07
Fix EADDRINUSE error in dev

### LINKED ISSUE #608: Race condition with server side reloading
There is a race condition when making changes to server side code which can cause an EADDRINUSE error.

### Type of issue

Bug

### Description

There is a race condition when making changes to server side code which can result in an EADDRINUSE error when the old process has not exited yet, and a new process is started.

### Steps to reproduce

1. scaffold a fusion app
2. add the following snippet into `src/main.js`

```js
if (__NODE__) {
  process.on('SIGTERM', () => {
    console.log('Received KIll');
    setTimeout(() => {
      process.exit();
    }, 10000);
  });
}
```

3. save the `src/main.js` file a few times

lhorie:
Fixes #608 

Depending on what's running on the server, a file save can trigger a EADDRINUSE error. This PR prevents the error from occurring by restarting the underlying server w/ a new port each time (thus allowing for the occupied port to be released in parallel).

There's a related [yarn issue](https://github.com/yarnpkg/yarn/issues/6608) that causes spawned subprocesses to become orphaned. To fully mitigate this issue, avoid upgrading to yarn 1.10.x or above.

### DIFF
--- a/build/dev-runtime.js
+++ b/build/dev-runtime.js
@@ -81,11 +81,10 @@ module.exports.DevelopmentRuntime = function(
     server: null,
     proc: null,
     proxy: null,
-    childPortP: getPort(),
   };
 
   this.run = async function reloadProc() {
-    const childPort = await state.childPortP;
+    const childPort = await getPort();
     const command = `
       const fs = require('fs');
       const path = require('path');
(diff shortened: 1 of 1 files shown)


lhorie:
re: yarn version, we can only recommend downgrading since the version of yarn is outside of our control

conradreuter:
For people using `patch-package` that seek a temporary fix without downgrading yarn:

Put the following file under `patches/fusion-cli+1.12.1.patch` and run `yarn`
https://gist.github.com/conradreuter/bcf252310a7ce848705d0f6a45f59b30

PR #612, opened 2018-11-08
Switch to parser-inserted scripts, fix script preloading for legacy bundles

rtsao:
**Summary**
- Fixes script preloading in legacy browsers
- Replaces dynamic script loading with scripts in markup to ensure the scripts are marked as parser-inserted. Not all browsers respect `defer` attributes for scripts that are not marked as parser-inserted. This caused problems in FF/Edge. https://html.spec.whatwg.org/multipage/scripting.html

  - Replace runtime Edge UA check with server-side UA check (to avoid bugs with native classes)
  - Replace runtime `nomodule` support check with server-side UA check for Safari 10.1 (to prevent double script download/execution)

### DIFF
--- a/plugins/ssr-plugin.js
+++ b/plugins/ssr-plugin.js
@@ -83,6 +83,7 @@ const SSRBodyTemplate = createPlugin({
 
       const legacyUrls = [];
       const modernUrls = [];
+
       for (let chunkId of allCriticalChunkIds) {
         const url = chunks.get(chunkId);
         if (url.includes('client-legacy')) {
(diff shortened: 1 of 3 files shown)


rtsao:
!merge

plugins/ssr-plugin.js
```diff
@@ -92,18 +93,61 @@ const SSRBodyTemplate = createPlugin({
...
+          );
+        }
+      }
+
+      const isSafari10_1 =
+        (browser.name === 'Safari' || browser.name === 'Mobile Safari') &&
+        browser.version.startsWith('10.1');
+
+      // Safari 10.1 supports modules but not `nomodule` attribute.
+      if (!isSafari10_1) {
```
angus-c: Might be reading this wrong but it looks like non-edge, non-safari 10.1 browsers get both modern and legacy bundles in `criticalChunkScripts`

rtsao: Yep, that's exactly as intended.

Browsers that do not support `<script type="module">` will simply ignore them. And all browser that support `<script type="module">` will ignore `<script nomodule>` scripts, *except* for Safari 10.1, which will download and execute both `type="module"` *and* `nomodule`. So exclusively for Safari 10.1, only `<script type="module">` should be sent.

angus-c: 👍

PR #614, opened 2018-11-09
Ensure credentials are sent for same-origin module chunks

rtsao:
Adds `crossorigin="anonymous"` attributes to `<script type="module">` scripts to ensure that credentials are sent for same-origin requests. Unlike regular scripts, credentials are not sent by default to same-origin sources.


rtsao:
!merge

PR #624, opened 2018-11-27
Update base jest config to handle nested package.json files

wentsul:
This PR updates the `base-jest-config` to work in a monorepo where there may be more than one package.json file. It first attempts to read the package file in the `cwd` and falls back to the package file in the git root directory.

This works for the following setup:

`my-service/package.json` has a `test` script that looks like:

`fusion test --configPath ../../node_modules/fusion-cli/build/jest/jest-config.js`

And the monorepo structure is as follows:

```
monorepo
├── package.json
└── services
    └── my-service
        └── package.json
```

### DIFF
--- a/build/jest/base-jest-config.js
+++ b/build/jest/base-jest-config.js
@@ -8,30 +8,43 @@
 
 /* eslint-env node */
 
+const child_process = require('child_process');
+
 const matchField = process.env.TEST_REGEX ? 'testRegex' : 'testMatch';
 const matchValue = process.env.TEST_FOLDER
   ? [`**/${process.env.TEST_FOLDER || '__tests__'}/**/*.js`]
   : process.env.TEST_REGEX ||
     (process.env.TEST_MATCH || '**/__tests__/**/*.js').split(',');
 
-function getReactVersion() {
-  // $FlowFixMe
-  const meta = require(process.cwd() + '/package.json');
-  const react =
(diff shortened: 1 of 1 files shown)


build/jest/base-jest-config.js
```diff
@@ -8,30 +8,43 @@
...
-  const react =
-    (meta.dependencies && meta.dependencies.react) ||
-    (meta.devDependencies && meta.devDependencies.react) ||
-    (meta.peerDependencies && meta.peerDependencies.react);
-  return react
-    .split('.')
-    .shift()
-    .match(/\d+/);
+function getGitRootDirectory() {
+  return child_process.execSync('git rev-parse --show-toplevel').toString().trim();
```
KevinGrandon: > git rev-parse --show-toplevel

We are moving towards having all CI commands run within docker containers, and we typically don't have git inside a container. We should enable what we need here with explicit CLI arguments instead. I thought that the `--dir` argument should do what you need?

wentsul: That makes sense. It looks like `--dir` sets the root directory when spawning the runner process which causes some issues with plugin loading (I'm seeing some translation failure issues and it'll output the coverage folder in the root directory which would cause issues with multiple projects).

I could just set `react` as a `peerDependency` in the local package and call it a day. Otherwise, it would be pretty simple to recursively check parent directories till a package.json file is found that has react as a dep. Let me know if you'd like me to make that update.

Issue #627, opened 2018-11-28
Fusion 1.13.0-beta.14 breaks non-strict mode code


vicapow:
To reproduce:

````sh
yarn create fusion-app common-js-bug-app
cd common-js-bug-app
````

Create a commonJS dependency.
````diff
diff --git a/src/common-js.js b/src/common-js.js
new file mode 100644
index 0000000..3a22e92
--- /dev/null
+++ b/src/common-js.js
@@ -0,0 +1,6 @@
+
+Foo = function() {
+  return 'common module';
+}
+
+module.exports.commonModule = () => Foo();
\ No newline at end of file
diff --git a/src/pages/home.js b/src/pages/home.js
index 40b3d85..dd28e86 100644
--- a/src/pages/home.js
+++ b/src/pages/home.js
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react';
 import {styled} from 'fusion-plugin-styletron-react';
+import {commonModule} from '../common-js';

 const Center = styled('div', {
   fontFamily: 'HelveticaNeue-Light, Arial',
@@ -63,7 +64,7 @@ const Home = () => (
       <Center>
         <Circle>
           <GettingStartedLink href="https://fusionjs.com/docs/getting-started">
-            Let&apos;s Get Started
+            {commonModule()}
           </GettingStartedLink>
         </Circle>
       </Center>
````

See that it runs normally with `yarn dev`. Then, update fusion-cli to the latest beta:

````diff
   "dependencies": {
-    "fusion-cli": "^1.12.1",
+    "fusion-cli": "1.13.0-beta.14",
````

See that it produces the following error:

````js
common-js.js:2 Uncaught ReferenceError: Foo is not defined
    at Object../src/common-js.js (common-js.js:2)
    at __webpack_require__ (bootstrap:782)
    at fn (bootstrap:150)
    at Module../src/pages/home.js (main.js:13)
    at __webpack_require__ (bootstrap:782)
    at fn (bootstrap:150)
    at Module../src/root.js (pageNotFound.js:11)
    at __webpack_require__ (bootstrap:782)
    at fn (bootstrap:150)
    at Module../src/main.js (main.js:2)
````

vicapow:
A workaround is to apply the following patch:

````
    {
      name: 'Dont use type=module for serving JS files.',
      details: '',
      file: './node_modules/fusion-cli/plugins/ssr-plugin.js',
      original: 'let forceLegacyOnly = false;',
[...]

Issue #622, opened 2018-11-15
ansi-color fails to bundle with 'Octal literal in strict mode' possibly due to fusion-cli's default babelification of node_modules 


lxe:
### Type of issue

Bug

### Description

If you import `ansi-color` package and attempt to `fusion {dev/build}` to bundle the project, the build fails with this error:

```
SyntaxError: /Users/aleksey/ansi-colors-app-test/node_modules/ansi-color/lib/ansi-color.js: Octal literal in strict mode (35:18)

  33 |   var ansi_str = "";
  34 |   for(var i=0, attr; attr = color_attrs[i]; i++) {
> 35 |     ansi_str += "\033[" + ANSI_CODES[attr] + "m";
     |                   ^
  36 |   }
  37 |   ansi_str += str + "\033[" + ANSI_CODES["off"] + "m";
  38 |   return ansi_str;
```

### Steps to reproduce

```
$ yarn create fusion-app ansi-colors-app-test
$ cd ansi-colors-app-test
$ yarn add ansi-color@0.2.1

$ patch -p1 <<EOF
diff --git a/src/main.js b/src/main.js
index b7729d6..d89fddb 100644
--- a/src/main.js
+++ b/src/main.js
@@ -5,0 +6 @@ import Styletron from 'fusion-plugin-styletron-react';
+import 'ansi-color'
EOF

$ yarn build
```

### Your environment

* fusion-cli version: 1.9.0

* Node.js version (`node --version`): v8.12.0

* npm version (`npm --version`): yarn 1.10.1

* Operating System: Mac

lxe:
@KevinGrandon it's empty. See repro steps. This happens in vanilla scaffold.

lhorie:
Yep, easily reproducible. Maybe a misconfiguration? https://github.com/babel/babel/issues/5240

vicapow:
curious if there was any update on this? for now we have to patch fusion using the following mechanism so it doesn't add the `use_strict` directive. This is an issue impacting more than just this module. Any module published to npm that relies on `non-use_strict` behavior is impacted.

vicapow:
````
    {
      name: `Fix D3, protobuf and other modules that should not be prefixed with 'use strict;'`,
      details: '',
      file: './node_modules/fusion-cli/build/compiler.js',
      original: `
            /node_modules\\/react-dom\\//,
            /node_modules\\/react\\//,
            /node_modules\\/core-js\\//,`,
      patch: `
[...]

PR #635, opened 2018-11-30
Replace module scripts with useragent checks

rtsao:
- Solves issues with Safari and same-origin credentials w/ module scripts:  https://bugs.webkit.org/show_bug.cgi?id=171550
- Fixes regression if bundles include code that must be run in non-strict mode(modules are always strict mode).

Fixes https://github.com/fusionjs/fusion-cli/issues/627
Fixes https://github.com/fusionjs/fusion-cli/issues/622


rtsao:
!merge

PR #641, opened 2018-12-03
Serve legacy bundles to Safari 10.1 and 11

rtsao:
Supersedes https://github.com/fusionjs/fusion-cli/pull/640

This solution is better because 10.1 and 11 are uncommon versions and this fix doesn't degrade the fast path for compliant browsers.


rtsao:
!merge

PR #643, opened 2018-12-04
Collect translations from translate function

akre54:
It's a pain keeping the list of translation key strings up to date with what's actually in use in the component. It'd be a lot cleaner if we automatically parsed calls to translate to collect strings.

Since `translationIds` is a Set, there's no downside to adding the same string multiple times (both from the `withTranslations` first array argument, or to multiple calls to `this.props.translate`.

Upsides of this pull implementation:
* Lot less manual management
* No risk of forgotten strings causing translation string to silently show up in UI -- currently no warning if `translate` can't find the string in the dictionary
* Only checks the `refPath` scope for `withTranslations` so will never translate a file that doesn't import `withTranslations`. 

Downsides:
* Translate function can't be reassigned (i.e. `const t = this.props.translate`, or `this.renderSomething({t: translate})`). Could be fixed if we walked the scope hierarchy or kept state.
* Doesn't resolve where translate function came from, so could incorrectly parse another translate function. This could be fixed if we use `TranslateType` flowtype as a hint.

### DIFF
--- a/build/babel-plugins/babel-plugin-i18n/index.js
+++ b/build/babel-plugins/babel-plugin-i18n/index.js
@@ -21,32 +21,12 @@ type PluginOpts = {translationIds: Set<string>}
 
 function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
   const t /*: Object */ = babel.types;
-  const moduleVisitor = createModuleVisitor(
+  const visitor = createModuleVisitor(
     t,
     COMPONENT_IDENTIFIER,
     PACKAGE_NAME,
     refsHandler
   );
-
-  const propVisitor = {
-    CallExpression(path, state) {
-      const {callee} = path.node;
-
-      if (callee.name === 'translate' || (callee.property && callee.property.name === 'translate')) {
-
-        const errorMessage = 'The translate function must be called with a string literal';
-
-        const element = path.node.arguments[0];
-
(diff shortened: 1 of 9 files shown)


ganemone:
I certainly agree that the `withTranslations` API could be improved, but I'm not sure this is quite the right way to improve it. I think the downsides you listed are too significant to go this route, but we could maybe do something similar to this.

One potential solution would be using a [custom hook](https://reactjs.org/docs/hooks-custom.html). This would resolve most of the static analysis issues, and would help us place more guards around usage. For example:

```js
import {useTranslation} from 'fusion-plugin-i18n-react';

function Component() {
  const translatedString = useTranslation('translation-key', {/* interpolation data */});
  return <div>{translatedString}</div>
}
```

akre54:
What about instead of a hook for each translate call, `useTranslation` returns a translate function that works much as the one now does? That way we can still walk the scope for `translate`, or `t`, or whatever you name the return value.

The issue with that API as I see it (without going down bikeshed lane too much) is that most places I see call translate inline (i.e. `const message = something ? translate('has.something') : translate('doesnt')`) , which would break one of the rules of hooks.

ganemone:
You are correct about the issue regarding access to context. Regarding the exact API of the translation hook, we should compare / contrast a few different options in a separate issue. I think your idea of returning a translate function is a good option. I'm going to close this PR for now, and open an issue describing the problem.

PR #647, opened 2018-12-11
Refactor tests, implement parallelized workers on CI

rtsao:
- Migrate tests to Jest
- Implements parallel containerized builds w/ Buildkite. Reduces CI test time from ~9 minutes to <3 minutes.


build/dev-runtime.js
```diff
@@ -86,6 +86,8 @@ module.exports.DevelopmentRuntime = function(
   this.run = async function reloadProc() {
     const childPort = await getPort();
     const command = `
+      process.on('SIGTERM', () => process.exit());
```
KevinGrandon: Why is this needed now, but not before? Just looking for some additional context.

test/e2e/assets-client-fetch/fixture/src/home.js
```diff
@@ -1,4 +1,4 @@
-// @flow
+// @noflow
```
KevinGrandon: Why noflow? Can we $FlowFixMe instead? (Same for other noflow changes in next few files)

rtsao:
Perhaps that's because Jest isn't showing the passing assertions (it only shows passing test files and failing assertions).

build/dev-runtime.js
rtsao: Without this, a zombie process was left each time dev was run.

PR #662, opened 2019-01-15
Add styletron-react displayName babel transform

chasestarr:
[link](https://github.com/styletron/styletron/tree/master/packages/babel-plugin-transform-styletron-display-name) to package readme.
When creating a component with `styletron-react`, it does not have a `displayName`. This leads to unclear component names in react dev tools.

### DIFF
--- a/build/get-babel-config.js
+++ b/build/get-babel-config.js
@@ -146,6 +146,7 @@ function fusionPreset(
       require.resolve('./babel-plugins/babel-plugin-sw'),
       require.resolve('./babel-plugins/babel-plugin-sync-chunk-paths'),
       require.resolve('./babel-plugins/babel-plugin-chunkid'),
+      require.resolve('babel-plugin-transform-styletron-display-name'),
       [
         require.resolve('babel-plugin-transform-cup-globals'),
         {target: targetEnv},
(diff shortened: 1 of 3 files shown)


build/get-babel-config.js
```diff
@@ -146,6 +146,7 @@ function fusionPreset(
       require.resolve('./babel-plugins/babel-plugin-sw'),
       require.resolve('./babel-plugins/babel-plugin-sync-chunk-paths'),
       require.resolve('./babel-plugins/babel-plugin-chunkid'),
+      require.resolve('babel-plugin-transform-styletron-display-name'),
```
KevinGrandon: Thanks for the PR! Wondering if there is two options that we could potentially do here:

1 - Ship this babel transform as part of the scaffold, included in the .fusionrc file.
2 - Ship this as a compiler plugin in fusion-plugin-styletron-react.

Just thinking that we should decouple concerns here if possible. Any thoughts @rtsao?

chasestarr: thanks for the review folks. I honestly need to get a better sense of the fusion plugin patterns.

FYI I added a new pr to the plugin [here](https://github.com/styletron/styletron/pull/290) so that it solves my original problem.

chasestarr:
closing, this was included in https://github.com/fusionjs/fusion-cli/pull/702

PR #664, opened 2019-01-16
Serve modern bundles to Chromium

rtsao:
The user agent parser identifies Chromium as a different browser, but it should be served modern bundles.


rtsao:
!merge

PR #666, opened 2019-01-17
Remove duplicated worker index

KevinGrandon:
Buildkite has recently added a worker index/total display to the UI so we no longer need to output this ourselves.

You can see the new index display here:

image


KevinGrandon:
!merge

PR #672, opened 2019-01-23
Fix browser field resolution in dependencies of sw bundle when es2015/es2017 fields are present

rtsao:
Vanilla browser field was working before, but not when es2015/es2017 fields were present.

This adds a fix and regression test.


rtsao:
!merge

PR #680, opened 2019-01-29
Pass timestamp as unused argument so sw.js updated for each build

angus-c:


build/loaders/sw-loader.js
```diff
@@ -86,5 +86,8 @@ function getCompiler(opts) {
 function getSWTemplateFnSource(swBundle) {
   return `export const swTemplate = (...params) => "${JSON.stringify(
     swBundle.toString()
-  ).slice(1, -1)};${SW_IDENTIFIER}(..." + JSON.stringify(params) + ")"`;
+  ).slice(
+    1,
+    -1
+  )};${SW_IDENTIFIER}(..." + JSON.stringify(params) + ",${Date.now()})"`;
```
rtsao: I wonder if instead this should be the hash of the sw bundle so that if the code is actually the same, then no unnecessary invalidation occurs.

angus-c: Won't that only change if referenced resource names change? In which case you don't need an extra check.

angus-c: Also the invalidation is on a resource by resource basis. We don't wipe out the whole cache when a new SW is installed. Just html and resources that changed

PR #697, opened 2019-02-08
Add module id dynamic import instrumentation

rtsao:
This is will enable [future changes to fusion-react](https://github.com/fusionjs/fusion-react/pull/227), including
- Removal of client-side `prepare` step
- Replacing custom React tree traversal algorithm in `prepare` with `renderToStaticMarkup` (which will enable hook support)

In particular, adding this instrumentation will enable the component returned by `split` to synchronously check if the module being dynamically imported has already been loaded (i.e. because it was preloaded on SSR) and then synchronously import it and render the component.


rtsao:
!merge

PR #699, opened 2019-02-12
Improve tests

lhorie:
Fix gql test regressions
Migrate i18n integration tests to fusion-cli


lhorie:
Yeah, I'm moving them out of fusion-plugin-i18n-react to remove a cyclical dependency between fusion-plugin-i18n-react and fusion-cli. This is needed to unblock putting these two packages in a monorepo

lhorie:
Note: I found another integration test in the other test file in fusion-plugin-i18n-react, so I moved it into here as well.

lhorie:
Yeah, I think we need to revisit the testing strategy here at some point. There's a bunch of stuff that probably should live in other packages, and also a bunch of tests that leave fixture files in weird states when tests throw or are aborted. Some refactoring targeting those issues will probably allow us to parallelize tests across CPUs to improve test times as well.

For now, since the tests are merely migrated from one repo to the other, release verification will eventually catch issues and this migration is low risk because we haven't done many changes to fusion-plugin-i18n-react lately anyways.

PR #702, opened 2019-02-14
Add babel-plugin-transform-styletron-display-name

ganemone:
Adds the styletron babel plugin which improves developer experience


ganemone:
!merge

PR #705, opened 2019-02-19
Add option in fusionConfig to configure jsExtPattern

Monar:
This might be minimalistic change allowing to experiment or develop with typescript?


KevinGrandon:
Thank you very much for the pull request. Digging into figure out what's going on with the Fusion.js compiler is no easy task 😅

I'm going to leave this open for the time being while I run it by our team. We currently don't use TS internally (yet), but definitely want to help support the community. My suggestion for the time being would be to fork this package and publish it to a registry. I'll try to keep this thread updated with the direction we want to go, or how we're going to support TS in the future.

Monar:
Good to hear your are open to enable typescript :) Publishing it separately sounds like a good idea for time being :)

Monar:
@KevinGrandon Hi I have strange observation regarding publishing this package, when I published it as `fusion-cli-ts` then `__DEV__` variable was not injected. I don't know where its injected but for sure not in this repository. My pocking at this problem fond relation of this `__DEV__` variable and package name what was quite unexpected by me... package name must end with `fusion-cli` whether it is `@scope/fusion-cli` or `whatever-fusion-cli`... both work but `fusion-cli-ts` is not ?!

KevinGrandon:
That is truly boggling since we should just inject the `__DEV__` variable through a babel plugin, which I don't believe relies on any package name.

It should be this line which registers the plugin: https://github.com/fusionjs/fusion-cli/blob/master/build/get-babel-config.js#L149

What is the error message you're getting? Is it undefined?

clayne11:
I just tried this same thing this morning and had also modify the `resolve.extensions` array to include `.ts` and `.tsx`. Were you able to get this working without doing that?

Monar:
Resolve extensions are for paths to packages without extensions file, in my testing so far I provided exact path to files so this didn't hit me :D I'll adjust it. 

And regarding the `__DEV__` and `fusino-cli-ts`, error is:
```bash
$ fusion dev
Using custom Babel config from .fusionrc.js
Warning: custom Babel config is an unstable API and may be not be supported in future releases. Use at your own risk.
(node:20282) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
Build completed in 3.313s

Build completed in 3.387s

ReferenceError: __DEV__ is not defined
    at reload (/***/node_modules/fusion-cli-ts/entries/server-entry.js:81:1)
    at process.runNextTicks [as _tickCallback] (internal/process/next_tick.js:47:5)
    at evalScript (internal/process/execution.js:61:11)
    at executeUserCode (internal/bootstrap/node.js:499:5)
    at startMainThreadExecution (internal/bootstrap/node.js:439:3)
```

I checked the code of this babel plugin you mentioned, but I have no experience in babel api so its hard to say is there a problem... but something somewhere is not right

PR #706, opened 2019-02-19
Add 'runInBand' to Jest option whitelist

lassedamgaard:
On our project we are seeing issues with integration test timeouts in CI. Running the tests sequentially solves the issues as described in https://jestjs.io/docs/en/troubleshooting#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server so having this option whitelisted would be extremely valuable.


KevinGrandon:
!merge

PR #709, opened 2019-02-20
Apply graphql-tag/loader on graphql/gql files

ganemone:
We need to apply this loader rather than use the loader alias workaround
because the loader relies on the ability to recursively load more graphql
files when importing fragments. This still works with the gql macro approach
and that can remain the recommended approach.


test/e2e/gql/fixture/src/query.gql
```diff
@@ -1,4 +1,4 @@
-# import './fragment.gql'
```
KevinGrandon: Was this working before?

ganemone: No - the test was broken also. I fixed the test and make it more robust to catch this kind of failure

ganemone:
!merge

PR #725, opened 2019-03-05
Fix preloading when chunk group contains multiple chunks

rtsao:
In some instances, chunks would not be preloaded when they should.

This would occur in a scenario like the following:

```js
// main.js
import("./a.js");
import("./b.js");
```

```js
// a.js
import shared from "./shared.js";
export default "a";
```

```js
// b.js
import shared from "./shared.js";
export default "b";
```

```js
// shared.js
export default "shared";
```

This would yield the following chunks (simplified naming)
- `chunk-runtime.js`
- `chunk-vendor.js`
- `chunk-main.js`
- `chunk-a.js`
- `chunk-b.js`
- `chunk-shared.js`

In webpack terminology, these chunks fit into three chunk groups:
- main group
  - `chunk-runtime.js`
  - `chunk-vendor.js`
  - `chunk-main.js`
- a group
  - `chunk-a.js`
  - `chunk-shared.js`
- b group
  - `chunk-b.js`
  - `chunk-shared.js`

**The bug**

`chunk-shared.js` would not be preloaded because the `import()`s for `a.js` and `b.js` would only be instrumented with the chunk ids for the `chunk-a.js` and `chunk-b.js` chunks, respectively.

A side effect of this meant that any translations in `shared.js` would be SSR'd, but would be missing on the initial client-side render because these translations would not be inlined into the SSR.

This PR ensures that `chunk-shared.js` is also added to the import promise instrumentation, ensuring that `chunk-shared.js` is preloaded correctly along with either `chunk-a.js` or `chunk-b.js`


test/e2e/dynamic-import-app-group-preloading/fixture/src/split-shared-ab.js
```diff
@@ -0,0 +1,977 @@
...
+import React from 'react';
+
+export default () => (
+  <div>
+    <span>split-shared-ab</span>
+    {reallyBigTextSoThisFileHasALargeFilesize}
+  </div>
+);
+
+var reallyBigTextSoThisFileHasALargeFilesize = `
```
KevinGrandon: Just leaving a comment here so when webpack changes down the road future generations have a place to laugh at us. Sorry! 😅

rtsao:
!merge

PR #729, opened 2019-03-14
Fix --no-hmr flag

micburks:
Pass hmr option through to compiler to prevent hmr plugins/middleware from being added to dev server


rtsao:
!merge

PR #739, opened 2019-03-19
Adding file list to package.json

madhusudhand:
fusion-cli npm modules includes all the files from source which makes it heavy.
Hence included only required files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,17 @@
     "lint": "eslint .",
     "test": "git clean -Xfd test/fixtures && jest --maxWorkers 1"
   },
+  "files": [
+    "bin/",
+    "build/",
+    "commands/",
+    "entries/",
+    "lib/",
+    "plugins/",
+    "get-compilation-metadata.js",
+    "README.md",
+    "LICENSE"
+  ],
   "dependencies": {
     "@babel/core": "^7.1.6",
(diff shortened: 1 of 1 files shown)


madhusudhand:
@KevinGrandon added a generic negate rule `"!**/test/"` to ignore all the test directories under any directory not just 'build'. In case if we add test files for others later this will take care.

package.json
```diff
@@ -10,6 +10,18 @@
     "lint": "eslint .",
     "test": "git clean -Xfd test/fixtures && jest --maxWorkers 1"
   },
+  "files": [
+    "README.md",
```
rtsao: This is not necessary per: https://docs.npmjs.com/files/package.json#files

```suggestion
```

package.json
```diff
@@ -10,6 +10,18 @@
...
+    "LICENSE",
```
rtsao: Also not necessary per: https://docs.npmjs.com/files/package.json#files

```suggestion
```

rtsao:
I think `sw.js` is also missing. This leaves me thinking maybe we might want to use `.npmignore` instead, which would be the contents of .gitignore plus:

```
docs/
test/e2e
.github/
.buildkite/
.*
```

I'm wary of accidentally having necessary files being published, so perhaps it's better to instead explicitly say which files should *not* be published.

PR #745, opened 2019-03-20
Use webpack native json behavior for loading json files via assetUrl

micburks:


rtsao:
Unless I'm mistaken, I think we should be able to remove `json-loader.js` entirely since this is built into webpack 4?

micburks:
@rtsao I thought so too, but removing it entirely doesn't seem to work with json files that were added to the build via assetUrl

rtsao:
Ah, I see, it must be taking precedence over file-loader?

micburks:
The comment in the json-loader also mentioned this would be a breaking change, but I haven't been able to identify how this breaks for a consumer. The only effect should be that unused keys should be removed from the client-main build. Am I missing something?

rtsao:
After comparing webpack v3 and v4 behavior, I think you are right -- there is no breaking change. I recall looking into this earlier and there was some different behavior regarding default exports, but maybe I was mistaken or this has changed in the meantime.

test/e2e/assets/test.js
```diff
@@ -89,11 +89,25 @@ test('`fusion dev` works with assets', async () => {
     );
     t.equal(await request(`${url}/dirname`), 'src');
     t.equal(await request(`${url}/filename`), 'src/main.js');
+
+    const jsonAssetUrl = await request(`${url}/json`);
+    const jsonAsset = (await request(url + jsonAssetUrl)).replace(/\s/g, '');
```
rtsao: Nitpick: maybe use `JSON.parse` and `t.deepEqual` instead?

micburks: Much better. Thanks!

micburks:
!merge

PR #750, opened 2019-03-25
Fix issue with errors being swallowed in child compilations

ganemone:
This ensures we get error logs whenever a compilation fails. It also adds a regression test
in the form of snapshots of the logger calls.


build/compiler.js
```diff
@@ -26,15 +26,39 @@ const {
 const mergeChunkMetadata = require('./merge-chunk-metadata');
 const loadFusionRC = require('./load-fusionrc.js');
 
+function getErrors(info) {
+  let errors = [].concat(info.errors);
+  if (info.children.length) {
```
AlexMSmithCA: Ah, good catch!

test/e2e/missing-module/test.js
```diff
@@ -3,6 +3,7 @@
 
 const path = require('path');
 const testEnvs = require('../errors.js');
+const t = require('assert');
```
AlexMSmithCA: (Minor) linter is complaining that this is never used:

{code}
$ eslint .
--
  | Warning: React version not specified in eslint-plugin-react settings. See https://github.com/yannickcr/eslint-plugin-react#configuration .
  | { parser: "babylon" } is deprecated; we now treat it as { parser: "babel" }.
  |  
  | /fusion-cli/test/e2e/missing-module/test.js
  | 6:7  error  't' is assigned a value but never used  no-unused-vars
  |  
  | ✖ 1 problem (1 error, 0 warnings)


{code}

ganemone:
!merge

PR #752, opened 2019-04-03
Fix EventEmitter leak in dev-runtime

cdlewis:
Lifecycle registered two event listeners using `once`. Only one would
unregister itself, leading to us leaking emitters. This could
potentially bring down the Fusion server as calls to `lifecycle.wait()`
would hang.

### DIFF
--- a/build/dev-runtime.js
+++ b/build/dev-runtime.js
@@ -27,6 +27,7 @@ function Lifecycle() {
   return {
     start: () => {
       state.started = true;
+      state.error = undefined;
       emitter.emit('message');
     },
     stop: () => {
@@ -48,6 +49,7 @@ function Lifecycle() {
           listening = true;
           emitter.once('message', (error /*: Error */) => {
             if (error) {
+              listening = false;
               return reject(error);
             }
 
--- a/build/dev-runtime.js
+++ b/build/dev-runtime.js
@@ -27,18 +27,16 @@ function Lifecycle() {
   return {
     start: () => {
       state.started = true;
-      state.error = undefined;
-      emitter.emit('started');
+      emitter.emit('message');
     },
     stop: () => {
       state.started = false;
     },
     error: error => {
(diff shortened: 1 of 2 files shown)


build/dev-runtime.js
```diff
@@ -27,18 +27,16 @@ function Lifecycle() {
...
       state.started = true;
-      state.error = undefined;
-      emitter.emit('started');
+      emitter.emit('message');
     },
     stop: () => {
       state.started = false;
     },
     error: error => {
-      state.error = error;
```
micburks: I'm not super familiar with the order these lifecycle methods are called, but wouldn't we still want `state.error = error` here so if `lifecycle.error` is called before `lifecycle.wait`, the promise returned from `wait` will reject instead of registering a `message` listener?

cdlewis: Yeah I'm not 100% sure either. I can change it back.

micburks:
!merge

test/e2e/server-startup-control/test.js
```diff
@@ -17,7 +17,7 @@ test('`fusion dev` proxy gracefully recovers from cached SSR errors', async () =
   async function waitForCompileToStart() {
     // Once the Fusion.js application renders SSR error pages in the app
     // we should leverage module.hot.addStatusHandler.
-    await new Promise(resolve => setTimeout(resolve, 3000));
+    await new Promise(resolve => setTimeout(resolve, 10000));
```
KevinGrandon: Is this still necessary? Seems this is not a timeout, and  adjusting this has a slight performance impact on tests. (Hopefully  in the future we change this so we don't need hard-coded number at all)

cdlewis: Yeah this was probably unnecessary. I put up a PR to undo https://github.com/fusionjs/fusion-cli/pull/757

PR #762, opened 2019-04-19
Update base jest config to ignore __generated__ directory when collecting coverage

ganemone:
Many code generation tools follow this pattern, so it makes sense to ignore this
file pattern by default.


ganemone:
!merge

PR #764, opened 2019-04-23
crossorigin script attribute needs to be set for CDN urls in order to support uncaught error handling

wentsul:
Couple things to note:

* This needs to be set for `__webpack_public_path__` but not sure if it should always be set for "critical chunk" scripts.
* I this fork deployed for a production and I'm now seeing client-side errors. Previously I was seeing "Script error." line 0.
* More information on why here: https://blog.sentry.io/2016/05/17/what-is-script-error
* Would to keep the tests but wasn't sure how to update the tests to only check for critical scripts. Changing `el.crossOrigin === 'anonymous'` was failing so I'm assuming there are other scripts.


test/e2e/dynamic-import-app/test.js
```diff
@@ -175,13 +175,6 @@ test('`fusion build` app with dynamic imports integration', async () => {
     'one extra script after loading new route'
   );
 
-  t.ok(
```
rtsao: The reason this crossorigin logic exists is because of:
https://github.com/webpack/webpack/issues/6972#issue-311819636

So the test is verifying that when not using a CDN, no crossorigin attributes are set. I don't think we want to delete these tests. I think this test should remain unchanged.

Instead, there *are* tests that does use a CDN, and we should verify that crossorigin attributes *are* set. As you point out, these assertions would probably currently fail.

plugins/ssr-plugin.js
```diff
@@ -114,8 +114,8 @@ const SSRBodyTemplate = createPlugin/*:: <SSRBodyTemplateDepsType,SSRBodyTemplat
 
         for (let url of criticalChunkUrls) {
           const crossoriginAttr = url.startsWith(__webpack_public_path__)
```
rtsao: Now think that this condition might not be the correct thing to check against, since `__webpack_public_path__` could be a CDN. Perhaps it should be `url.startsWith("/")`, but I suppose that doesn't handle the case where the absolute url is same-origin (although this case might not matter).

wentsul: It looks like the [legacy ssr template](https://github.com/fusionjs/fusion-core/blob/master/src/plugins/ssr.js#L139) used `webpackPublicPath.startsWith('http')` which is similar to what you mentioned. What about checking  for`CDN_URL`?

```js
const crossoriginAttr = process.env.CDN_URL ? ' crossorigin="anonymous"' : '';
```

test/e2e/dynamic-import-app/test.js
wentsul: added tests for CDN case if we want to go that route.

PR #766, opened 2019-05-01
Update enzyme and enzyme-adapter-react-16 to latest

thegleb:
We'd like to use `React.memo` in a project that uses `fusion-cli`; support for `React.memo` was added to `enzyme-adapter-react-16` fairly recently: https://github.com/airbnb/enzyme/pull/1914/files

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "case-sensitive-paths-webpack-plugin": "^2.1.2",
     "chalk": "^2.4.1",
     "compression-webpack-plugin": "^2.0.0",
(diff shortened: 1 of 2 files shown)


thegleb:
Not super sure how to add the appropriate label, which is probably `dependencies`...

KevinGrandon:
Tests passed so we are good to go here, typically this would go out with our next release, but for now we can publish a pre-release so you can test this.

KevinGrandon:
!merge

thegleb:
Thanks @KevinGrandon ! Looks like the pre-release works as I was hoping.

PR #769, opened 2019-05-03
Add translation key instrumentation for dynamic imports

micburks:
Annotate dynamic imports with the translation keys the chunk contains

Fixes issues with dynamically loading translations where routing isn't guaranteed to hit the origin the initial request was served from. This will allow looking up a translation by key instead of the id of the chunk that contains it.

Essentially I had to split the i18nManifest state into its `DeferredState` and the `Map` that it resolves to. This way, the server build can wait for the client build to finish (as before) by referencing the `DeferredState` and the instrumentation plugin can use the `Map` directly to lookup translation keys.

This is supplemented by PRs in `fusion-react`, `fusion-plugin-i18n`, and `fusion-plugin-i18n-react`.

### DIFF
--- a/build/get-webpack-config.js
+++ b/build/get-webpack-config.js
@@ -65,6 +65,7 @@ const JS_EXT_PATTERN = /\.jsx?$/;
 /*::
 import type {
   ClientChunkMetadataState,
+  TranslationsManifest,
   TranslationsManifestState,
   LegacyBuildEnabledState,
 } from "./types.js";
@@ -86,7 +87,7 @@ export type WebpackConfigOpts = {|
     clientChunkMetadata: ClientChunkMetadataState,
     legacyClientChunkMetadata: ClientChunkMetadataState,
     mergedClientChunkMetadata: ClientChunkMetadataState,
-    i18nManifest: Map<string, Set<string>>,
+    i18nManifest: TranslationManifest,
     i18nDeferredManifest: TranslationsManifestState,
     legacyBuildEnabled: LegacyBuildEnabledState,
   },
--- a/build/get-webpack-config.js
+++ b/build/get-webpack-config.js
@@ -472,18 +472,19 @@ function getWebpackConfig(opts /*: WebpackConfigOpts */) {
       watch && new webpack.NoEmitOnErrorsPlugin(),
       runtime === 'server'
         ? // Server
-          new InstrumentedImportDependencyTemplatePlugin(
-            state.mergedClientChunkMetadata
-          )
+          new InstrumentedImportDependencyTemplatePlugin({
+            compilation: 'server',
+            clientChunkMetadata: state.mergedClientChunkMetadata,
+          })
         : /**
            * Client
            * Don't wait for the client manifest on the client.
            * The underlying plugin is able determine client chunk metadata on its own.
            */
-          new InstrumentedImportDependencyTemplatePlugin(
-            void 0,
-            state.i18nManifest
-          ),
+          new InstrumentedImportDependencyTemplatePlugin({
+            compilation: 'client',
+            i18nManifest: state.i18nManifest,
+          }),
       dev && hmr && watch && new webpack.HotModuleReplacementPlugin(),
       !dev && runtime === 'client' && new webpack.HashedModuleIdsPlugin(),
       runtime === 'client' &&
@@ -535,10 +536,10 @@ function getWebpackConfig(opts /*: WebpackConfigOpts */) {
               options.optimization.splitChunks
             ),
             // need to re-apply template
-            new InstrumentedImportDependencyTemplatePlugin(
-              void 0,
-              state.i18nManifest
-            ),
+            new InstrumentedImportDependencyTemplatePlugin({
+              compilation: 'client',
+              i18nManifest: state.i18nManifest,
+            }),
             new ClientChunkMetadataStateHydratorPlugin(
               state.legacyClientChunkMetadata
             ),
--- a/build/get-webpack-config.js
+++ b/build/get-webpack-config.js
@@ -478,7 +478,7 @@ function getWebpackConfig(opts /*: WebpackConfigOpts */) {
         : /**
            * Client
            * Don't wait for the client manifest on the client.
-           * The underlying plugin handles client instrumentation on its own.
+           * The underlying plugin is able determine client chunk metadata on its own.
            */
           new InstrumentedImportDependencyTemplatePlugin(
             void 0,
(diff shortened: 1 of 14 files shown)


micburks:
> PR title is invalid: First word must be imperative verb.

`Instrument` can be an imperative verb

build/plugins/instrumented-import-dependency-template-plugin.js
```diff
@@ -69,13 +78,45 @@ class InstrumentedImportDependencyTemplate extends ImportDependencyTemplate {
       chunkIds = getChunkGroupIds(depBlock.chunkGroup);
     }
 
+    let translationKeys = [];
+    if (this.translationsManifest) {
+      const modulesSet = new Set();
+
+      // Module dependencies
+      if (dep.module && dep.module.dependencies) {
```
rtsao: Interesting, so these modules in `module.dependencies` are not already in `chunk._modules` for each chunk in the chunk group?

Regardless, I think it might be worth encapsulating logic this into a separate function:
`dep => Set<string>`.

lhorie (dismissed):

build/plugins/instrumented-import-dependency-template-plugin.js
```diff
@@ -69,13 +78,45 @@ class InstrumentedImportDependencyTemplate extends ImportDependencyTemplate {
...
+      }
+
+      // Chunks
+      depBlock.chunkGroup.chunks.forEach(chunk => {
+        const modules = Array.from(chunk._modules.keys());
+        modules.forEach(m => modulesSet.add(m.resource));
+      });
+
+      const modules = Array.from(modulesSet.keys());
+      translationKeys = modules.reduce((acc, module) => {
```
rtsao: Because the `translationKeys` array is already instantiated in an outer lexical scope (which we assign to), I think using `reduce` is probably not providing much benefit. So barring a situation where we care about immutability, I think plain iteration and array mutation is slightly simpler (and less code).

```js
for (const module of modules) {
  if (translationsManifest.has(module)) {
    const keys = this.translationsManifest.get(module).keys();
    translationKeys.push(...keys);
  }
}
```

(I also removed the `Array.from`, since spread element works on any type of iterable)

build/get-webpack-config.js
```diff
@@ -465,17 +470,20 @@ function getWebpackConfig(opts /*: WebpackConfigOpts */) {
...
-             * Don't wait for the client manifest on the client.
-             * The underlying plugin handles client instrumentation on its own.
-             */
-            void 0
-      ),
+          )
+        : /**
+           * Client
+           * Don't wait for the client manifest on the client.
+           * The underlying plugin handles client instrumentation on its own.
```
rtsao: This is no longer accurate now that the i18n manifest comes from outside.
```suggestion
           * The underlying plugin is able determine client chunk metadata on its own.
```

build/plugins/instrumented-import-dependency-template-plugin.js
```diff
@@ -69,13 +78,45 @@ class InstrumentedImportDependencyTemplate extends ImportDependencyTemplate {
...
+        } else {
+          return acc;
+        }
+      }, []);
+    }
+
     // Add the following properties to the promise returned by import()
     // - `__CHUNK_IDS`: the webpack chunk ids for the dynamic import
     // - `__MODULE_ID`: the webpack module id of the dynamically imported module. Equivalent to require.resolveWeak(path)
+    // - `__I18N_KEYS`: the translation keys that are used in this bundle
```
rtsao:
```suggestion
    // - `__I18N_KEYS`: the translation keys used in the client chunk group for this import() 
```

build/plugins/instrumented-import-dependency-template-plugin.js
micburks: Agreed

build/plugins/instrumented-import-dependency-template-plugin.js
micburks: I originally didn't have to use `module.dependencies`, but I learned through testing that `chunk._modules` wasn't working in a production build for the `main` bundle. I don't fully understand it, but `module.dependencies` works in production and `chunk._modules` works in development

micburks: I think I can elaborate on this a little more. In dev, `chunk._modules` only contains `NormalModule`s whereas in the production build most modules (especially components that import other components) are replaced by a few `ConcatenatedModule`s. In order to get the filenames from this I have to do a little digging into `dep.module.dependencies`.  This makes the set of filenames larger than it probably has to be, but it also seems to be accurate (i.e. contains files we need and not ones we don't need).

build/get-webpack-config.js
```diff
@@ -465,17 +470,20 @@ function getWebpackConfig(opts /*: WebpackConfigOpts */) {
...
-            void 0
-      ),
+          )
+        : /**
+           * Client
+           * Don't wait for the client manifest on the client.
+           * The underlying plugin is able determine client chunk metadata on its own.
+           */
+          new InstrumentedImportDependencyTemplatePlugin(
+            void 0,
```
rtsao: Now that this constructor now takes more than one parameter, I think an object might make this more understandable. This was already a bit confusing (my bad) where the argument is expected to be undefined in the client case, but server-side should be provided. Adding another parameter now makes it even worse.

Expressing the parameter type as a disjoint union of two different objects would help I think. For example:
```js
type Opts = 
  | ClientOpts
  | ServerOpts;

type ServerOpts = {
  compilation: "server",
  clientChunkMetadata: ClientChunkMetadataState
};

type ClientOpts = {
  compilation: "client",
  i18nManifest: TranslationsManifest
};
```

Also, this makes me think: is it odd that `__I18N_KEYS` is only added on the client-side? The other two properties are added in both the server and client bundles.

I think in the future, potentially we could align the server code to work more like the client code. I think this particular promise instrumentation might be useful even in a Suspense-powered i18n system.

rtsao (changes requested):

build/get-webpack-config.js
micburks: Yeah I think that'll make the plugin easier to understand. I'll update to use those types.

micburks: I was actually thinking the opposite for the promise instrumentation. I don't totally understand why `__CHUNK_IDS` are added to the server build since they're only used when dynamically loading chunks with new translations (at least I think that's the only time they're used).

rtsao: `__CHUNK_IDS` is necessary server-side because that's how the server knows which async chunks are used in the SSR and are thus considered "critical" and should therefore be preloaded. During SSR, when it encounters a `split` component, it adds the chunk ids (on the promise object) to the list of critical chunks that should be preloaded.

micburks: Okay I recall that now. Since I didn't change how the translation manifest is added to the server bundle, the server still looks up the necessary translations based on the `__CHUNK_IDS` of the promise. We could potentially change that to use use `__I18N_KEYS` but yes probably not really necessary at this point. It might require a different approach than what I have because this approach doesn't find any translation keys since they're all bundled together in the server build.

PR #771, opened 2019-05-06
Add a tree shaking test that involves the use of Flow types

alxmyth:
As part of the investigation into looking into [WPT-2994](https://jeng.uberinternal.com/browse/WPT-2994) this PR adds a new test to ensure that `imports`/`exports` that include Flow types are properly stripped and tree shaken.


test/e2e/tree-shaking-flow/fixture/src/main.js
```diff
@@ -0,0 +1,59 @@
...
+  //  - userland
+  const unustedUserlandVal: UnusedUserlandType = { shouldBeUnused: unusedUserlandValue };
+  const unustedUserlandVal2: UnusedUserlandOtherType = { alsoShouldBeUnused: unusedUserlandValue };
+
+  console.log(unusedDepVal, unusedDepVal2, unustedUserlandVal, unustedUserlandVal2);
+}
+
+// used in the browser
+//  - node modules
+const usedDepVal: UsedDependencyType = { shouldBeUsed: usedDependencyValue};
```
alxmyth, line 51: Lines `51-52` here cause the vendor bundle to leak the unused import from `fixture-pckg` (`__FIXTURE_DEPENDENCY_UNUSED__` which should be used only in the server bundle):

```js
/*! no static exports found */function(e,o,s){"use strict";Object.defineProperty(o,"__esModule",{value:!0});o.usedDependencyValue="__FIXTURE_DEPENDENCY_USED__",o.unusedDependencyValue="__FIXTURE_DEPENDENCY_UNUSED__"}}]);
```

Removing these two lines tree shakes the import for `fixture-pckg`.

test/e2e/tree-shaking-flow/fixture/node_modules/fixture-pkg/dist/index.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
```
alxmyth, line 1: We couldn't have `export type` declarations in the dependency's entry point, so maybe this test isn't the best...

Currently using the transpiled `index.js` here and added an `index.js.flow` file for Flow/IDE integration.

alxmyth:
I was having trouble reproducing whether code was leaking into the browser due to Flow imports:

e.g.
```js
import type {AType} from ... // no leak
import {B} from ...
```
vs.
```js
import {type AType, B} from ... // potential leak
```

Not entirely sure why, but tree splitting seems to be failing for node module dependency imports as a separate concern.

